PR #5, opened 2024-09-14
[feat]: generate knowledge files automatically

brandonkachen:
Tests passing locally:
Screenshot 2024-09-17 at 12 07 36 AM

meme: 
image

### DIFF
--- a/test/__mock-data__/hallucinated/expected.ts
+++ b/test/__mock-data__/hallucinated/expected.ts
@@ -12,7 +12,7 @@ import {
 } from 'common/util/file'
 import { getSystemPrompt } from './system-prompt'
 import { STOP_MARKER } from 'common/constants'
-import { getTools } from 'common/src/util/tools'
+import { getTools } from './tools'
 import { Message } from 'common/actions'
 import { ToolCall } from 'common/actions'
 import { debugLog } from './util/debug'
@@ -21,11 +21,11 @@ import { generatePatch } from './generate-patch'
 import { requestRelevantFiles } from './request-files-prompt'
 
 function processChunk(chunk: string): string {
-  const fileBlockRegex = /<file[\s\S]*?<\/file>/g
+  const fileBlockRegex = /<file[\s\S]*?<\/file>/g;
   return chunk.replace(fileBlockRegex, (match) => {
-    const firstLine = match.split('\n')[0]
-    return `${firstLine}\n...\n</file>`
-  })
+    const firstLine = match.split('\n')[0];
+    return `${firstLine}\n...\n</file>`;
+  });
 }
 
 async function streamResponse(
@@ -33,37 +33,29 @@ async function streamResponse(
   ws: WebSocket,
   messageId: string
 ): Promise<string> {
-  let fullResponse = ''
-  let currentChunk = ''
+  let fullResponse = '';
+  let currentChunk = '';
 
   for await (const chunk of response) {
-    const content = chunk.choices[0]?.delta?.content || ''
-    fullResponse += content
-    currentChunk += content
+    const content = chunk.choices[0]?.delta?.content || '';
+    fullResponse += content;
+    currentChunk += content;
 
     // Process the chunk when we have a complete sentence or a significant amount of text
-    if (
-      content.includes('.') ||
-      content.includes('\n') ||
-      currentChunk.length > 100
-    ) {
-      const processedChunk = processChunk(currentChunk)
-      ws.send(
-        JSON.stringify({ type: 'chunk', content: processedChunk, messageId })
-      )
-      currentChunk = ''
+    if (content.includes('.') || content.includes('\n') || currentChunk.length > 100) {
+      const processedChunk = processChunk(currentChunk);
+      ws.send(JSON.stringify({ type: 'chunk', content: processedChunk, messageId }));
+      currentChunk = '';
     }
   }
 
   // Send any remaining content
   if (currentChunk) {
-    const processedChunk = processChunk(currentChunk)
-    ws.send(
-      JSON.stringify({ type: 'chunk', content: processedChunk, messageId })
-    )
+    const processedChunk = processChunk(currentChunk);
+    ws.send(JSON.stringify({ type: 'chunk', content: processedChunk, messageId }));
   }
 
-  return fullResponse
+  return fullResponse;
 }
 
 /**
@@ -144,14 +136,7 @@ ${STOP_MARKER}
       const fileBlocks = parseFileBlocks(currentFileBlock)
       for (const [filePath, newFileContent] of Object.entries(fileBlocks)) {
         fileProcessingPromises.push(
-          processFileBlock(
-            userId,
-            ws,
-            messages,
-            fullResponse,
-            filePath,
-            newFileContent
-          )
+          processFileBlock(userId, ws, messages, fullResponse, filePath, newFileContent)
         )
 
         currentFileBlock = currentFileBlock.replace(fileRegex, '')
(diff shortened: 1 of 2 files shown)


backend/src/generate-knowledge-files.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  console.log('knowledge file messages:', messages)
+  const response = await promptClaude(messages, {
+    userId,
+    system: systemPrompt,
+    tools: DEFAULT_TOOLS,
+  })
+  const fileContentMatch = response.match(/<file[^>]*>([\s\S]*)<\/file>/)
+  const fileContent = fileContentMatch ? fileContentMatch[1].trim() : ''
+  const filePath = 'knowledge.md' // TODO: get path to best location for knowledge file, use `fileContext.currentWorkingDirectory`?
```
jahooma: You can use `parseFileBlocks` to get all the files updated. File blocks also include the filepath. The type is an object with the key as the filepath and the value as the file content.

So you can allow it to choose the knowledge files to change in the prompt and it could even pick multiple. In your run it seemed like it wanted to update a particular knowledge file within npm-app.

backend/src/generate-knowledge-files.ts
```diff
@@ -0,0 +1,78 @@
...
+  const userPrompt = `
+    Here are some relevant files and code diffs that you should consider: 
+    ${getRelevantFilesPrompt(fileContext)}
+    
+    If the change isn't important enough to warrant a new knowledge file, please do not output anything. We don't want to waste the user's time on irrelevant changes.
+    This is also meant to be helpful for future LLMs like yourself. Thus, please be concise and avoid unnecessary details. If the change is important, please provide a detailed description of what we're doing and why.
+    Again, this is  meant to be useful for both humans and LLMs. 
+
+    If you determined that the change is important enough to warrant a knowledge file, please see the following instructions on how to create a helpful knowledge file:
+    ${knowledgeFilesPrompt}
```
jahooma: The relevant files prompt and knowledge files prompt seem like information that would be in the system prompt? 

The user prompt could use chain of thought to ask to summarize the current conversation first, and then ask it if there's anything important to add to a knoweldge file, and if so, create file blocks with those changes.

jahooma: You might also want to include the edit files prompt in the system prompt so it can competently create the file blocks

backend/src/main-prompt.ts
```diff
@@ -49,6 +53,18 @@ export async function mainPrompt(
...
+
+    // Generate new/updated knowledge files if there are signficant changes to document
+    const knowledgeFiles = await generateKnowledgeFiles(
+      userId,
+      ws,
+      fullResponse,
+      fileContext,
+      messages
+    )
+    fileProcessingPromises.push(...knowledgeFiles)
```
jahooma: I think awaiting generateKnowledgeFields means it will block manicode's normal generation until finished. We don't want to await

backend/src/generate-knowledge-files.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export async function generateKnowledgeFiles(
+  userId: string,
+  ws: WebSocket,
+  fullResponse: string,
+  fileContext: ProjectFileContext,
+  initialMessages: Message[]
+): Promise<Promise<FileChange>[]> {
+  // Prompt Claude to check for some heuristics around whether or not we've done enough meaningful work to warrant creating a knowledge file.
+  // TODO: take into account the actual diffs the user manually made to the files – these likely represent little issues that we didn't do correctly
```
jahooma: BTW, I'm now including user edits within the user message itself, so it will be included automatically within the messages

backend/src/main-prompt.ts
```diff
@@ -49,6 +57,29 @@ export async function mainPrompt(
...
     fullResponse += responseChunk
     shouldCheckFiles = false
+  } else {
+    // Already have context from existing chat
+
+    // If client used tool, we don't want to generate knowledge files because the user isn't really in control
+    const clientUsedTool = match(lastMessage)
+      .with(
+        {
+          role: 'user',
```
brandonkachen: based on this pattern: 

Screenshot 2024-09-16 at 10 16 12 PM

jahooma: Looks good. I think this could be helpful as a util function: isToolResult. Could create a new file within common for it

backend/src/main-prompt.ts
```diff
@@ -183,20 +212,14 @@ ${STOP_MARKER}
...
 
-  const changeAppendix =
-    changes.length > 0
-      ? `\n\n<edits_made_by_assistant>\n${changes
-          .map(({ filePath, content }) => createFileBlock(filePath, content))
-          .join('\n')}\n</edits_made_by_assistant>`
-      : ''
-
   return {
-    response: `${fullResponse}${changeAppendix}`,
```
jahooma: Oh, I don't think we want to remove this!

npm-app/knowledge.md
```diff
@@ -53,6 +53,8 @@ The build process is defined in `package.json`:
 
 ## Dependencies
 
+- ts-pattern: Used for pattern matching in TypeScript. Installed to improve type safety and readability when checking message types.
```
jahooma, line 56: Nice! This is a cool library.

test/__src__/generate-knowledge-files.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  expectedFiles: FileChange[]
+) => {
+  const responses = await generateKnowledgeFiles(
+    'userId',
+    mockWs,
+    '',
+    fileContext,
+    messages
+  )
+  expect(responses.length).toBeLessThanOrEqual(expectedFiles.length)
```
jahooma: Wouldn't we want to check if the lengths are the same? Otherwise, it could be not generating any knowledge files.

A slightly better test might be to await the file blocks, and check the file paths that were edited to see if they include the right knowledge file

npm-app/package.json
```diff
@@ -52,6 +52,7 @@
     "tree-sitter-ruby": "0.23.0",
     "tree-sitter-rust": "0.23.0",
     "tree-sitter-typescript": "0.23.0",
+    "ts-pattern": "^5.3.1",
```
jahooma: For production dependencies, I like to not use "^" in the version because occasionally a new version can cause problems. It's not a huge deal though

backend/src/main-prompt.ts
brandonkachen: named it `didClientUseTool`, since i'm also checking for the role to be `user`.

test/__src__/generate-knowledge-files.test.ts
brandonkachen: Good point, there are really two scenarios going on here: 
1. the LLM might not output a file where we expect one.
2. the LLM might output more files than we expect.

IMO case 1 is not ok, since knowledge isn't captured, but case 2 is ok since this means Claude decides to create a `web-scraper.knowledge.md` instead of `knowledge.md`. (If we want it to always update `knowledge.md`, then I can tighten the test and add more prompting to help guide it to always write to a `knowledge.md` file.)

If you're in agreement, I can add code for these two cases above.

backend/src/main-prompt.ts
brandonkachen: Fixed in latest commit

test/__src__/generate-knowledge-files.test.ts
brandonkachen: Discussed and decided to just add a bit of extra code to handle these two cases.

PR #16, opened 2024-10-04
[feat] implement API limits for anon, free, and paid users

brandonkachen:


test/__src__/quota-manager.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      set: mock(() => ({
+        where: mock(() => ({
+          execute: mock(),
+        })),
+      })),
+    })),
+  },
+}))
+
+describe('QuotaManager', () => {
```
jahooma, line 36: This is great, test are good for us

brandonkachen: lovingly coded by manicode 😌

PR #17, opened 2024-10-09
[feat] referrals system

brandonkachen:
https://www.tella.tv/video/referral-program-explained-7qdh

**important:** when we deploy, make sure to change  APP_URL to NEXT_PUBLIC_APP_URL in our environment variables!


common/src/db/schema.knowledge.md
```diff
@@ -0,0 +1,88 @@
+# Referral Feature Implementation Plan
```
brandonkachen, line 1: will prob delete this before merging

PR #19, opened 2024-10-13
[fix] cli show usage better

brandonkachen:
Screenshot 2024-10-12 at 6 56 51 PM

demo: https://www.tella.tv/video/fixing-cli-usage-limits-bd2q


backend/src/websockets/middleware.ts
```diff
@@ -82,9 +80,6 @@ protec.use(async (action, _clientSessionId, _) => {
   logger.debug(`Protecting action of type: '${action.type}'`)
 })
 protec.use(async (action, _clientSessionId, ws) => {
-  if (env.ENVIRONMENT === 'local') {
```
brandonkachen, line 85: removed due to convo about consolidating environment checks (open to adding back in the future if that proves impossible)

PR #24, opened 2024-10-26
[fix] use stored fingerprint for logged in users

brandonkachen:
https://www.tella.tv/video/cm2pobk97000003l487io0d98

### DIFF
--- a/npm-app/src/client.ts
+++ b/npm-app/src/client.ts
@@ -48,7 +48,6 @@ export class Client {
     this.returnControlToUser = returnControlToUser
   }
 
-  // can make public
   private async getFingerprintId(): Promise<string> {
     if (this.fingerprintId) {
       return this.fingerprintId
--- a/npm-app/src/fingerprint.ts
+++ b/npm-app/src/fingerprint.ts
@@ -11,91 +11,51 @@ import {
   // @ts-ignore
 } from 'systeminformation'
 
-// export const FINGERPRINTING_INFO = (async function () {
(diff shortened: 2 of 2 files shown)


npm-app/src/index.ts
```diff
@@ -4,31 +4,28 @@ import fs from 'fs'
...
 
-  const readyPromise = Promise.all([
-    updatePromise,
-    fingerprintPromise,
-    initFileContextPromise,
-  ])
+  const readyPromise = Promise.all([updatePromise, initFileContextPromise])
 
-  const cli = new CLI(readyPromise, { autoGit })
+  const defaultFingerprintId = await fingerprintPromise
```
jahooma: Hmm, does this delay the startup at all? Seems like it might

brandonkachen: just tested, it's ~250ms on avg + another ~250ms to get the actual device info – but that led me to figuring out a nice speed-up boost for logged-in users!

we were getting the device info on every app load (because the `FINGERPRINT_INFO` const is actually an anon fn invocation that is runs each time the app starts). so by turning it into a normal fn and only calling that when we need to calculate fingerprint (i.e. for non-logged-in users), we can avoid that unnecessary work for normal users! 

might not actually save a full 500ms, since we're calling these functions asynchronously at start-up, but it _feels_ a bit faster to me :D and definitely not blocking the. main thread, like my previous attempt

jahooma: Nice! That's a great improvement

PR #29, opened 2024-11-05
[feat] rename to Codebuff

brandonkachen:


web/src/app/page.tsx
```diff
@@ -73,7 +73,7 @@ const Home = () => {
               <p>Backed by</p>
               <Link
                 target="_blank"
-                href="https://www.ycombinator.com/companies/manicode"
+                href="https://www.ycombinator.com/companies/codebuff"
```
brandonkachen, line 76: oh shoot we'll need to update our co name in the YC directory

PR #31, opened 2024-11-06
making the codebase work on Windows

jannalulu:
webpage renders but backend still doesn't work. Also still creating new migration files in common/src instead of using existing files in common/src/db

### DIFF
--- a/common/src/db/schema.knowledge.md
+++ b/common/src/db/schema.knowledge.md
@@ -20,19 +20,7 @@ Important: The database setup requires:
    - Run schema operations
 
 Note: Setup has been primarily tested on Mac. Windows users may encounter platform-specific issues:
-- Use forward slashes in schema paths (convert using path.replace(/\\/g, '/'))
-- Avoid using Unix-specific commands like 'sleep' in npm scripts
 - When using __dirname or path.join() in config files, convert Windows backslashes to forward slashes
-- Module imports may need explicit /dist path on Windows (e.g., @t3-oss/env-nextjs/dist)
-- Always test path resolution in both development and production environments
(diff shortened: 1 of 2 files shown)


common/src/db/drizzle.config.ts
```diff
@@ -4,8 +4,8 @@ import path from 'path'
 
 export default defineConfig({
   dialect: 'postgresql',
-  schema: path.join('src', 'db', 'schema.ts'),
-  out: path.join('src', 'db', 'migrations'),
+  schema: path.join(__dirname, 'schema.ts').replace(/\\/g, '/'),
+  out: './src/migrations',
```
jahooma: This line looks suspicious

common/src/db/env.mjs
```diff
@@ -1,17 +1,29 @@
...
+console.log('Loading stack.env from:', stackEnvPath)
+const stackResult = dotenv.config({ path: stackEnvPath })
+console.log('Stack env loaded:', stackResult)
+
+// Force set the environment from stack.env
+if (stackResult.parsed?.ENVIRONMENT) {
+  process.env.ENVIRONMENT = stackResult.parsed.ENVIRONMENT
+}
+
+console.log('ENVIRONMENT after loading stack.env:', process.env.ENVIRONMENT)
```
jahooma: Remove console.logs before merging

jannalulu: removed

common/src/db/drizzle.config.ts
brandonkachen: @jannalulu can you try 'src/db/migrations' here instead?

jannalulu: done

PR #36, opened 2024-11-08
edit tsconfig.json and package.json files

jannalulu:
windows doesn't like "./" and wants to use "../"

node as a peerDependency was wrong according to codebuff

added missing packages to dependencies in package.json ("ignore" and "ts-pattern") <-- in hindsight this should have been a separate commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,13 +26,6 @@
     "deploy:npm-app": "bun run bump && cd npm-app && npm publish",
     "test:patch": "bun test test/__src__/patch.test.ts"
   },
-  "dependencies": {
-    "@tanstack/react-query": "^5.59.16",
-    "ignore": "^6.0.2",
-    "lodash": "4.17.21",
-    "ts-node": "^10.9.2",
-    "ts-pattern": "^5.5.0"
-  },
   "devDependencies": {
     "@types/lodash": "4.17.7",
     "@types/node": "^22.9.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,15 +28,18 @@
   },
   "dependencies": {
```
jahooma: I don't see why we need any dependencies for the root package.json. It is not a project that builds code directly. We only need dev dependencies

test/tsconfig.json
```diff
@@ -19,7 +19,8 @@
       "common/*": ["../common/src/*"],
       "code-map/*": ["../packages/code-map/*"]
     },
-    "types": ["bun-types"]
+    "types": ["bun-types"],
+    "outDir": "./dist"
```
jahooma, line 23: We don't need an out dir, because we are not compiling anything. This just runs tests using typescript

package.json
jannalulu: not sure; "@tanstack/react-query and lodash were already there.

test/tsconfig.json
jannalulu: Windows throws `Cannot write file '../test/__src__/env.mjs' because it would overwrite input file.` if I don't have outDir.

jahooma: Ah, ok. Seems harmless I guess

package.json
jahooma: Remove them, it should work

jannalulu: done

PR #37, opened 2024-11-08
more renaming

jannalulu:
more renaming random things


web/src/lib/constant.ts
```diff
@@ -6,6 +6,7 @@ export const siteConfig = {
     'Codebuff is a tool for editing codebases via natural language instruction to Mani, an expert AI programming assistant.',
```
jahooma: One more reference to Mani here

PR #43, opened 2024-11-19
[feat] Send usage data to Stripe

brandonkachen:


backend/src/billing/message-cost-tracker.ts
```diff
@@ -97,4 +101,57 @@ export const saveMessage = async (value: {
     credits: creditsUsed,
     finished_at: value.finishedAt,
   })
+
```
brandonkachen, line 104: Commenting out logs because they would add a lot of noise in most circumstances.

PR #75, opened 2025-01-25
[feat] auto browse with puppeteer

brandonkachen:
https://www.tella.tv/video/troubleshooting-browser-issues-7d3i

### DIFF
--- a/backend/src/deepseek-api.ts
+++ b/backend/src/deepseek-api.ts
@@ -197,7 +197,7 @@ export const promptDeepseekStream = (
     clientSessionId: string
     fingerprintId: string
     userInputId: string
-    model: string
+    model: DeepseekModel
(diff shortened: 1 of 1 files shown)


common/src/actions.ts
```diff
@@ -156,6 +157,15 @@ export const ResponseCompleteSchema = z
   )
 
 export const SERVER_ACTION_SCHEMA = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal('browser-action'),
+    instruction: BrowserActionSchema,
+    xml: z.string().optional(),
+  }),
+  z.object({
+    type: z.literal('read-files-response'),
```
brandonkachen, line 173: Was this never in here? Did Codebuff hallucinate this addition, or were we missing an important change? o.o

PR #93, opened 2025-03-16
[fix] show usage credits on each message

brandonkachen:
The new version puts the unsubscribe complete back into its original place and instead as the backend send a duplicate of the original usage data to the client and then the client will track what it's received so far and only incremented counter once for a given request.

So now we have accurate usage updates because the client has multiple paths to receive the usage costs and is idempotent to multiple usage requests. However, this means that tool call chains will result in _each_ individual usage showing up with the request, which may mislead people into thinking these are cheaper than they are:
Screenshot 2025-03-15 at 8 30 55 PM

### DIFF
--- a/npm-app/src/client.ts
+++ b/npm-app/src/client.ts
@@ -384,7 +384,6 @@ export class Client {
       if (!parsedAction.success) return
 
       this.setUsage(parsedAction.data)
-      this.showUsageWarning()
     })
   }
 
@@ -652,6 +651,8 @@ export class Client {
           )
           this.hadFileChanges = false
         }
+
+        this.showUsageWarning()
 
         unsubscribeChunks()
         unsubscribeComplete()


npm-app/src/client.ts
```diff
@@ -576,22 +612,17 @@ export class Client {
...
-          // TODO: figure out when to unsubscribe
-          unsubscribeComplete()
-          return
+        if (action.promptId !== userInputId) return
+        
+        // Handle usage data if present
+        if (this.pendingRequestId === userInputId) {
+          const usageData = UsageReponseSchema.omit({ type: true }).safeParse(a)
+          if (usageData.success) {
+            this.processUsageResponse(usageData.data)
```
jahooma: I think we'd prefer to only print the usage once at the end of the userInputId's response (when isComplete = true).

I would decouple updating usage from printing about it, and not print here, just below, when isComplete is true. Then you don't need the instance variable this.usageProcessed, which is hard to reason about when we can have concurrent responses (since you can cancel one and start another while the cancelled one will still return).

My other suggestion is to track the usage total for the userInputId in a mutable variable local to the subscribeToResponse function. I'll add another comment.

npm-app/src/client.ts
```diff
@@ -536,6 +569,9 @@ export class Client {
       }
       setMessages(newMessages)
 
+      this.pendingRequestId = null
```
jahooma: You could add `let totalRequestUsage = 0` here. And print that value when isComplete=true. Much easier to reason about than instance variables IMO

npm-app/src/client.ts
```diff
@@ -601,30 +632,25 @@ export class Client {
...
-            // Save lastChanges for `diff` command
             this.lastChanges.push(FileChangeSchema.parse(toolCall.parameters))
-
             this.hadFileChanges = true
           }
           if (
             toolCall.name === 'run_terminal_command' &&
             toolCall.parameters.mode === 'user'
           ) {
-            // Special case: when terminal command is run it as a user command, then no need to reprompt assistant.
```
jahooma: Don't let codebuff delete comments haha

PR #94, opened 2025-03-18
Dynamically load tree sitter packages

jahooma:
TODO: throw an error and quit codebuff if they require sudo permissions. Print a message the at they should run `sudo codebuff`

### DIFF
--- a/npm-app/package.json
+++ b/npm-app/package.json
@@ -51,6 +51,9 @@
     "puppeteer-core": "^24.2.0",
     "systeminformation": "5.23.4",
     "tree-sitter": "^0.21.0",
+    "tree-sitter-javascript": "^0.23.0",
+    "tree-sitter-python": "^0.23.0",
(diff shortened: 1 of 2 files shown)


npm-app/package.json
```diff
@@ -50,22 +50,59 @@
...
+    "tree-sitter-c-sharp": "^0.23.0",
+    "tree-sitter-cpp": "^0.23.0",
+    "tree-sitter-go": "^0.23.1",
+    "tree-sitter-java": "^0.23.2",
+    "tree-sitter-javascript": "^0.23.0",
+    "tree-sitter-php": "^0.23.2",
+    "tree-sitter-python": "^0.23.2",
+    "tree-sitter-ruby": "^0.23.0",
+    "tree-sitter-rust": "^0.23.0",
+    "tree-sitter-typescript": "^0.23.0"
```
jahooma: Feel free to move some package to dependencies instead of peerDependencies if we want to always include them

PR #108, opened 2025-04-11
wip: feat: initial filepicker ft experiments

venkr:
Hacky finetuning demo PR. 
Not mergeable but helpful to pre-empt any unknown unknowns.

This PR corresponds to my "v0 task" described in the prev doc: https://www.notion.so/codebuff/Venki-s-model-Distilled-file-finding-1ce8ada876c38080bee1d7321434d9dd?pvs=4#1cf8ada876c380fda41cd9172950f7cb

**What it does:**
- Added DB tables (probably gonna change this to Clickhouse/Snowflake etc)
- Captures file-picker pairs
- Script to relabel outputs w/ Claude-3.5 and Gemini-2.5-pro
- Script to download finetuning data as JSONL for a given relabeler model
- Script to trigger a finetuning run (only tested OpenAI gpt-4o-mini for now, probably Gemini-2.0-flash is our main actual test!)

**Big things missing:**
- Storage we're long-run happy with -> will chat today 
- Finetuning Gemini-2.0-flash -> will get service keys today
- Capturing additional input data (ie: store input prompts in a structured way, so we can test variants on the backend. Maybe even finetune w/ a shorter prompt, for example)
- Capturing data for evals (ie: capture the _next_ main_prompt after the file retrieval, so we could see what claude-3.5 would see, and ask it to rate how good the files are) [TBD: Maybe weird bias in including some but not all full-files]

### DIFF
--- a/scripts/ft-file-selection/collect-tuning-data.ts
+++ b/scripts/ft-file-selection/collect-tuning-data.ts
@@ -1,4 +1,4 @@
-import { writeFileSync } from 'fs'
+import { existsSync, readdirSync, writeFileSync } from 'fs'
 
 import db from 'common/db'
 import { ft_filepicker_capture, ft_filepicker_traces } from 'common/db/schema'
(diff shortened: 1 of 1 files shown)


backend/src/find-files/request-files-prompt.ts
```diff
@@ -1,23 +1,26 @@
-import { WebSocket } from 'ws'
-import { range, shuffle, uniq } from 'lodash'
 import { dirname, isAbsolute, normalize } from 'path'
-import { TextBlockParam } from '@anthropic-ai/sdk/resources'
 
+import { TextBlockParam } from '@anthropic-ai/sdk/resources'
```
venkr, line 3: TODO: Fix my linter - it's reordering differently 🤔

PR #112, opened 2025-04-17
feat: add ai sdk for finetuned vertex models

venkr:
✅ Merged: This PR does NOT intentionally modify any production pathways yet, so should be very safe.
💭 Open to any feedback that i'll address in a follow-up since this will become the primary LLM interface though.

------

This PR:
- Adds Vercel AI SDK
- Adds a Vertex model for the test finetuned model (finetuned on like 2 samples, so likely not any good!)

I've successfully been able to run requests against the finetuned model.

Further goals of this unified API:
- Handle retries, fallbacks, timeouts, in a centralized place
- Avoid code duplication of eg: saveMessage() and time measurement code 
- Avoid needing to transform messages ourselves - rely on Vercel's CoreMessage + their setup to transform to other formats.

Ideally, the final interface will look something like:
```export const promptAiSdk = async function (
  messages: CoreMessage[],
  options: {
    clientSessionId: string
    fingerprintId: string
    userInputId: string
    model: Model | ModelConfig[] // Note: change
    userId: string | undefined
    maxTokens?: number
    temperature?: number
  }
): Promise<string> {
```

where
`ModelConfig` is of form `[{model: Model, retries?: number}, {model: Model, retries?: number}...]`
Perhaps ModelConfig would also eg have fields to specify: timeout, retry backoff duration etc. (Or maybe that'd live in options)

### DIFF
--- a/backend/src/llm-apis/gemini-with-fallbacks.ts
+++ b/backend/src/llm-apis/gemini-with-fallbacks.ts
@@ -74,7 +74,6 @@ export async function promptGeminiWithFallbacks(
       geminiOptions
     )
   } catch (error) {
-    console.log('error', error)
     logger.warn(
       { error },
       'Error calling Gemini API, falling back to Vertex Gemini'


backend/src/llm-apis/vercel-ai-sdk/ai-sdk.ts
```diff
@@ -0,0 +1,305 @@
...
+    finishedAt: new Date(),
+    latencyMs: Date.now() - startTime,
+  })
+
+  return content
+}
+
+// TODO: temporary - ideally we move to using CoreMessage[] directly
+// and don't need this transform!!
+function transformMessages(
```
venkr, line 132: This is a kinda yucky function.
We'd ideally move to directly using `CoreMessage` in the codebase, since ai-sdk will deal w/ the work of massaging it into provider specific formats for us.

PR #116, opened 2025-04-21
feat: file-picker finetuning

venkr:
Add BigQuery data collection + scripts to relabel & generate JSONL training data


common/src/bigquery/client.ts
```diff
@@ -0,0 +1,248 @@
...
+import {
+  GetRelevantFilesTrace,
+  Relabel,
+  RELABELS_SCHEMA,
+  Trace,
+  TRACES_SCHEMA,
+} from './schema'
+
+const DATASET =
+  process.env.ENVIRONMENT === 'production'
```
brandonkachen, line 13: do we want to store our locally-generated data into the `codebuff_data_dev` project? or is that reserved for backend staging/dev-type data?

common/src/bigquery/client.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.tracesTable = 'traces'
+    this.relabelsTable = 'relabels'
+  }
+
+  async initialize() {
+    try {
+      // Ensure dataset exists
+      const [dataset] = await this.client
+        .dataset(this.dataset)
+        .get({ autoCreate: true })
```
brandonkachen: Should probably disable this, in case we get into a bad edge case and accidentally create a ton of unnecessary datasets

common/src/bigquery/client.ts
```diff
@@ -0,0 +1,248 @@
...
+      SELECT * FROM ${this.dataset}.${this.tracesTable}
+      ORDER BY createdAt DESC
+      LIMIT ${limit}
+    `
+    const [rows] = await this.client.query(query)
+    // Parse the payload as JSON if it's a string
+    return rows.map((row) => ({
+      ...row,
+      payload:
+        typeof row.payload === 'string' ? JSON.parse(row.payload) : row.payload,
```
brandonkachen: is it possible for the row we get back from bigquery to be string or json only?

common/src/bigquery/client.ts
venkr: locally generated data goes there - and I think that's fine!

PR #198, opened 2025-07-03
[fix]: remove Anthropic fallback mechanism and use OpenRouter for all Claude models instead

brandonkachen:
- Remove anthropic-with-fallbacks.ts and related test files
- Replace all claudeModels references with OpenRouter equivalents
- Update cost tracking to include OpenRouter model pricing
- Simplify AI SDK integration by removing fallback logic
- Update provider checks to use 'openrouter' instead of 'anthropic'
- Consolidate all Claude model access through OpenRouter API

This change simplifies the codebase by using a single provider (OpenRouter) for all Claude models, removing the complexity of fallback mechanisms while maintaining full Claude model support.

Generated with Codebuff 🤖

### DIFF
--- a/evals/git-evals/gen-evals.ts
+++ b/evals/git-evals/gen-evals.ts
@@ -1,14 +1,20 @@
 import { execSync } from 'child_process'
 import fs from 'fs'
-import path from 'path'
 import { chunk } from 'lodash'
+import path from 'path'
 import { z } from 'zod'
 
-import { claudeModels, geminiModels } from '@codebuff/common/constants'
 import { promptAiSdkStructured } from '@codebuff/backend/llm-apis/vercel-ai-sdk/ai-sdk'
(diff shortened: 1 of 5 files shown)


backend/src/llm-apis/message-cost-tracker.ts
```diff
@@ -74,19 +73,18 @@ const TOKENS_COST_PER_M = {
...
-    [models.sonnet]: 3.75,
-    [models.sonnet3_7]: 3.75,
-    [models.haiku]: 1,
   },
   cache_read: {
-    [models.opus4]: 1.5,
-    [models.sonnet]: 0.3,
-    [models.sonnet3_7]: 0.3,
-    [models.haiku]: 0.08,
     [models.deepseekChat]: 0.014,
```
jahooma: Just flagging this deletion: don't we need to still be charging for cache creation and reads, even if via open router?

PR #219, opened 2025-07-25
[feat] initial agent store schema

brandonkachen:
- Added publisher table with slug, username, email, verified status, and bio
- Added agent table with version, generated major/minor/patch fields, and template config
- Used PostgreSQL generated columns to automatically parse semantic versions
- Simplified from complex multi-table design to flexible JSONB-based approach
- Template field stores all agent configuration as structured JSON

🤖 Generated with Codebuff

### DIFF
--- a/common/src/db/migrations/0022_whole_juggernaut.sql
+++ b/common/src/db/migrations/0022_whole_juggernaut.sql
@@ -1,0 +1,33 @@
+CREATE TABLE IF NOT EXISTS "agent_template" (
+	"id" text,
+	"version" text NOT NULL,
+	"publisher_id" text NOT NULL,
+	"major" integer GENERATED ALWAYS AS (CAST(SPLIT_PART("agent_template"."version", '.', 1) AS INTEGER)) STORED,
+	"minor" integer GENERATED ALWAYS AS (CAST(SPLIT_PART("agent_template"."version", '.', 2) AS INTEGER)) STORED,
+	"patch" integer GENERATED ALWAYS AS (CAST(SPLIT_PART("agent_template"."version", '.', 3) AS INTEGER)) STORED,
+	"template" jsonb NOT NULL,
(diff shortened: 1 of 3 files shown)


common/src/db/schema.ts
```diff
@@ -418,3 +418,51 @@ export const gitEvalResults = pgTable('git_eval_results', {
...
+  created_at: timestamp('created_at', { mode: 'date', withTimezone: true })
+    .notNull()
+    .defaultNow(),
+  updated_at: timestamp('updated_at', { mode: 'date', withTimezone: true })
+    .notNull()
+    .defaultNow(),
+})
+
+export const agent = pgTable(
+  'agent',
```
jahooma: consider agent_template

common/src/db/schema.ts
```diff
@@ -418,3 +418,51 @@ export const gitEvalResults = pgTable('git_eval_results', {
...
+  {
+    id: text('id')
+      .primaryKey()
+      .$defaultFn(() => crypto.randomUUID()),
+    publisher_id: text('publisher_id')
+      .notNull()
+      .references(() => publisher.id, { onDelete: 'cascade' }),
+    version: text('version').notNull(), // Semantic version e.g., '1.0.0'
```
jahooma: primaryKey() right?

common/src/db/schema.ts
```diff
@@ -418,3 +418,51 @@ export const gitEvalResults = pgTable('git_eval_results', {
...
     .defaultNow(),
 })
+
+// Agent Store tables
+export const publisher = pgTable('publisher', {
+  id: text('id')
+    .primaryKey()
+    .$defaultFn(() => crypto.randomUUID()),
+  slug: text('slug').unique().notNull(), // for store reference
+  username: text('username').unique().notNull(),
```
jahooma: Should this just be name? Do we need it to be unique?

PR #221, opened 2025-07-27
[feat][render preview][codecane] agent builder

brandonkachen:
TODOS:
- [x] Add yield 'STEP_ALL' (add a user message with context and instructions right before)
- [x] Selecting create new agent does questionare + invokes agent-builder in main view
- [x] Support /agents reseting conversation with selected agent
- [x] Option in /agents to edit agents (which could just directly invoke the agent-builder?)
- [x] test on render preview + codecane


backend/src/templates/agents/agent-builder.ts
```diff
@@ -129,9 +101,144 @@ Analyze their request and create a complete agent template that:
...
+        toolName: 'run_terminal_command',
+        args: {
+          command: 'mkdir -p .agents/templates',
+          process_type: 'SYNC',
+          timeout_seconds: 10,
+          cb_easp: true,
+        },
+      }
+
+      // Step 2: Write the agent-template.d.ts file with the template content
```
jahooma, line 188: Use a terminal command to get our hosted file from a url and save it locally

backend/src/templates/agents/agent-builder.ts
```diff
@@ -129,9 +101,144 @@ Analyze their request and create a complete agent template that:
...
+      }
+
+      // Step 1: Create directory structure
+      yield {
+        toolName: 'run_terminal_command',
+        args: {
+          command: 'mkdir -p .agents/templates',
+          process_type: 'SYNC',
+          timeout_seconds: 10,
+          cb_easp: true,
```
jahooma: [C] This should not be required

backend/src/run-programmatic-step.ts
```diff
@@ -143,7 +143,10 @@ export async function runProgrammaticStep(
 
   try {
     // Execute tools synchronously as the generator yields them
+    let stepCount = 0
     do {
+      stepCount++
```
jahooma: Did we want to log the stepCount? Otherwise remove

PR #254, opened 2025-08-31
feat(sdk): enable code_search in sdk

brandonkachen:
this was vibe coded heh

@jahooma pls take a look!

🤖 Generated with Codebuff


sdk/src/tools/code-search.ts
```diff
@@ -0,0 +1,80 @@
...
+}): Promise<CodebuffToolOutput<'code_search'>> {
+
+  return new Promise((resolve) => {
+    let stdout = ''
+    let stderr = ''
+
+    const flagsArray = (flags || '').split(' ').filter(Boolean)
+    let searchCwd = cwd
+    
+    // Note: In the SDK, we don't have access to a project root concept,
```
jahooma: The client object has a this.cwd, which is like the project root. The cwd from the tool call should be concatenated with that!

sdk/src/tools/code-search.ts
```diff
@@ -0,0 +1,80 @@
+import { spawn } from 'child_process'
+import * as path from 'path'
+
+import { rgPath } from '@vscode/ripgrep'
+
+import type { CodebuffToolOutput } from '../../../common/src/tools/list'
+
+export function codeSearch({
```
jahooma, line 8: We need to hook this up in `handleToolCall` inside `sdk/src/client.ts`

sdk/src/tools/run-file-change-hooks.ts
```diff
@@ -0,0 +1,21 @@
+import type { CodebuffToolOutput } from '../../../common/src/tools/list'
+
+export function runFileChangeHooks({
```
jahooma, line 3: Ditto for hooking this up in client.ts

bun.lock
```diff
@@ -232,8 +229,9 @@
     },
     "sdk": {
       "name": "@codebuff/sdk",
-      "version": "0.1.18",
+      "version": "0.1.19",
       "dependencies": {
+        "@vscode/ripgrep": "^1.15.9",
```
jahooma: Let's nail down the version for public deps (no "^")

PR #258, opened 2025-09-05
Fix Windows HEREDOC compatibility in git commit commands

PierrunoYT:
## Problem

Windows users encounter fatal errors when Codebuff generates git commit commands using HEREDOC syntax:

```
fatal: paths '<<'EOF' ...' with -a does not make sense
```

The issue occurs because HEREDOC syntax (`<<'EOF'`) only works in bash/Unix shells, not Windows Command Prompt.

## Solution

Updated system prompts to provide cross-platform compatible commit command examples:

- **Unix/bash shells**: Continue using HEREDOC syntax for multi-line commits
- **Windows Command Prompt**: Use simple quoted multi-line strings
- **PowerShell**: Use here-string syntax (`@"..."@`)

## Changes

1. **`backend/src/system-prompt/prompts.ts`**: Updated commit instruction examples to show platform-specific syntax with clear guidance on when to use each approach
2. **`backend/src/tools/definitions/tool/run-terminal-command.ts`**: Replaced HEREDOC example with Windows-compatible format

## Testing

The fix ensures commit commands work across all platforms while maintaining:
- Co-authored attribution
- Proper commit message formatting
- Multi-line commit support

🤖 Generated with Claude Code
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/backend/src/system-prompt/prompts.ts
+++ b/backend/src/system-prompt/prompts.ts
@@ -300,16 +300,6 @@ When the user requests a new git commit, please follow these steps closely:
    Co-Authored-By: Codebuff <noreply@codebuff.com>"
    \`\`\`
    
-   **For PowerShell:**
-   \`\`\`
-   git commit -m @"
-   Your commit message here.
-
-   🤖 Generated with Codebuff
-   Co-Authored-By: Codebuff <noreply@codebuff.com>
-   "@
-   \`\`\`
-   
    Always detect the platform and use the appropriate syntax. HEREDOC syntax (\`<<'EOF'\`) only works in bash/Unix shells and will fail on Windows Command Prompt.
 
 **Important details**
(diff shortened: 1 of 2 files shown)


backend/src/system-prompt/prompts.ts
```diff
@@ -289,13 +291,26 @@ When the user requests a new git commit, please follow these steps closely:
...
+   
+   **For PowerShell:**
+   \`\`\`
+   git commit -m @"
+   Your commit message here.
+
+   🤖 Generated with Codebuff
+   Co-Authored-By: Codebuff <noreply@codebuff.com>
+   "@
+   \`\`\`
```
jahooma: You can delete this Powershell section.

I believe in Windows we always run in cmd.exe and not powershell (even in powershell, we invoke cmd.exe to run commands)

jahooma:
Nice!! Glad you finally found a way to fix this issue -- after we open sourced haha.

Congrats on the first external PR! 🎉

PR #260, opened 2025-09-06
[render preview] Bundle ripgrep in common package to eliminate npm rate limiting issues.

brandonkachen:
This adds cross-platform ripgrep binaries to common, implements a resolver for Node and compiled modes, updates npm-app and sdk to use the bundled version, removes @vscode/ripgrep dependency, and includes a download script for maintaining binaries.

Generated with Codebuff 🤖


sdk/src/tools/code-search.ts
```diff
@@ -1,11 +1,11 @@
 import { spawn } from 'child_process'
 import * as path from 'path'
 
-import { rgPath } from '@vscode/ripgrep'
+import { getRgPath } from '@codebuff/common/native/ripgrep'
```
jahooma: We've been importing all common files with relative imports FYI. This only works if we publish @codebuff/common as another npm package

PR #305, opened 2025-09-21
Fix: Update agents page link to correct /store URL

thisisharsh7:
Fixed broken link in README.md - updated agents page link from /agents to /stores to point to the correct published agents store page.

Corrected link text from https://www.codebuff.com/agents to https://www.codebuff.com/store


jahooma:
Thanks @thisisharsh7!! Appreciate it

PR #312, opened 2025-09-24
[feat][render preview][codecane] benchify integration

brandonkachen:


backend/src/tools/batch-str-replace.ts
```diff
@@ -0,0 +1,969 @@
...
+export type BatchStrReplaceState = {
+  deferredStrReplaces: DeferredStrReplace[]
+  otherToolsQueue: any[]
+  strReplacePhaseComplete: boolean
+  failures: any[]
+}
+
+const BENCHIFY_FILE_TYPES = ['tsx', 'ts', 'jsx', 'js']
+const BENCHIFY_TIMEOUT_MS = 3000 // 3 second timeout for Benchify calls
+const BENCHIFY_MAX_FILES = 10 // Maximum files to send to Benchify
```
maxvonhippel, line 34: We actually could do quite a bit more than this potentially.  For one of our app-builder customers we often get requests with 70+ files, and Juan's deploying blob support currently to make this quite a bit more network-efficient.  So maybe worth revisiting.

backend/src/tools/batch-str-replace.ts
```diff
@@ -0,0 +1,969 @@
...
+    // Call Benchify with timeout and retry logic
+    const benchifyResult = await callBenchifyWithResilience(
+      filteredChanges,
+      options,
+    )
+
+    if (benchifyResult && benchifyResult.length > 0) {
+      logger.info(
+        {
+          benchifyResultCount: benchifyResult.length,
```
maxvonhippel, line 500: I don't think this makes sense to log. It's the length of a resolved string.

maxvonhippel: I would think this was added previously when we were giving you `ALL_FILES` before we moved to `DIFF`.

PR #345, opened 2025-10-18
Add sourcemap

dannyighsu:
For ease of debugging for SDK consumers.

Tested on local w/ the fork to confirm source maps appear and are viewable.

### DIFF
--- a/sdk/scripts/build.ts
+++ b/sdk/scripts/build.ts
@@ -43,7 +43,7 @@ async function build() {
     target: 'node',
     format: 'esm',
     minify: false,
-    sourcemap: 'external',
+    sourcemap: 'linked',
     external,
     naming: '[dir]/index.mjs',
     loader: {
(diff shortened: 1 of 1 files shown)


sdk/package.json
```diff
@@ -18,6 +18,7 @@
   "sideEffects": false,
   "files": [
     "dist",
+    "dist/**/*.map",
```
jahooma: I think this is unnecessary b/c the whole directory is included already

sdk/scripts/build.ts
```diff
@@ -57,6 +58,7 @@ async function build() {
     target: 'node',
     format: 'cjs',
     minify: false,
+    sourcemap: 'external',
```
jahooma: I wonder if the two source maps conflict? Maybe we should just generate one, IDK

jahooma:
Thanks Danny!

sdk/scripts/build.ts
dannyighsu: I believe you need both since consumers will be either/or and not both

PR #348, opened 2025-10-21
[fix] switch off lodash, appease dependabot

brandonkachen:


common/src/util/lodash-replacements.ts
```diff
@@ -0,0 +1,122 @@
+// Shared lodash replacement utilities
+// These functions replace lodash with native JavaScript implementations
+
+// Deep clone using JSON serialization (works for serializable objects)
+export function cloneDeep<T>(obj: T): T {
+  return JSON.parse(JSON.stringify(obj)) as T
+}
+
+// Deep equality check using JSON serialization
+export function isEqual(a: unknown, b: unknown): boolean {
```
charleslien: This implementation doesn't work for all cases:

`a = { hello: 1, world: 2 }`
`b = { world: 2, hello: 1 }`
Are equal, but the strings are not equal

Do you know how lodash implements it?

PR #385, opened 2025-11-27
Require JSON output for tools

charleslien:


codebuff-team:
The core idea — moving from the old union `ToolResultOutput` (`{type:'json',value}` vs `{type:'media',...}`) to a single flattened shape (`{value, media?}`) and renaming `$getToolCallString` → `$getNativeToolCallExampleString` to reflect native JSON tool calling instead of the `<codebuff_tool_call>` text convention — is a reasonable direction, and the mechanical rename across all `common/src/tools/params/tool/*.ts` description strings is done consistently.

However this PR bundles far too much into one changeset with an empty description:

1. **Unrelated agent logic rewrites**: `.agents/editor/best-of-n/editor-best-of-n.ts` and `.agents/file-explorer/file-picker.ts` get substantial rewrites of `extractSpawnResults`/`extractLastMessageText` plumbing that have nothing to do with the stated title. These changes assume a new spawn-result shape (`agentOutput.type === 'lastMessage'`) that isn't explained or justified anywhere in the diff.
2. **New integration tests that call live LLM APIs** (`editor-best-of-n.integration.test.ts`, `file-explorer.integration.test.ts`) require a real `CODEBUFF_API_KEY`, run for 60-120s, and assert loosely on substring matches in freeform model output. These are flaky and expensive to run in CI, and one of the three is already `.skip`'d with a comment admitting the underlying feature doesn't fully work yet.
3. **`cli/src/hooks/use-send-message.ts`** silently changes the condition for adding tool calls to subagent blocks from `if (agentId)` to `if (parentAgentId && agentId)` — a behavior change buried in a large diff with no explanation of why.
4. **Cost aggregation test assertions change** (73→80, 50→60 credits) without justification for why the new native tool-call format changes credit costs — this looks like tests adjusted to match new behavior rather than verifying it's correct.

Given the private repo receives hand-ported changes, this diff is too large and unfocused to port safely — the tool-output-shape change, the agent-logic rewrite, and the CLI behavior change should be three separate PRs, each with a description explaining the reasoning and edge cases handled.

PR #413, opened 2026-01-17
Require bash on Windows. Throw error if missing

jahooma:


82deutschmark:
I can second this!

jahooma:
Which terminal you use to start codebuff is separate from which shell codebuff uses to run commands. It should always run commands in bash if it can. This works even if you start codebuff in powershell

PR #423, opened 2026-01-28
Subscription backend

jahooma:


common/src/constants/subscription-plans.ts
```diff
@@ -0,0 +1,26 @@
+export const PLAN_NAMES = ['pro'] as const
```
brandonkachen: hmmm this should be derived from PLANS below

common/src/constants/subscription-plans.ts
```diff
@@ -0,0 +1,26 @@
...
+  name: PlanName
+  displayName: string
+  monthlyPrice: number
+  creditsPerBlock: number
+  blockDurationHours: number
+  weeklyCreditsLimit: number
+}
+
+export const PLANS = {
+  pro: {
```
brandonkachen: need a better name

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,335 @@
...
+    .where(eq(schema.user.stripe_customer_id, customerId))
+    .limit(1)
+  return userRecord[0]?.id ?? null
+}
+
+/**
+ * Resolves a PlanConfig from a Stripe price ID.
+ * Compares against the configured env var for each plan.
+ */
+function getPlanFromPriceId(priceId: string): PlanConfig {
```
brandonkachen: instead of needing functions like this, perhaps we can add/modify a const in common/src/constants/subscription-plans.ts to be an object whose keys are the subscription IDs set to metadata about them.

then we can simply check `SUBSCRIPTIONS[priceId]` for its info

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,335 @@
...
+import { handleSubscribe } from './subscription'
+
+import type { Logger } from '@codebuff/common/types/contracts/logger'
+import type { PlanConfig } from '@codebuff/common/constants/subscription-plans'
+import type Stripe from 'stripe'
+
+/**
+ * Looks up a user ID by Stripe customer ID.
+ */
+async function getUserIdByCustomerId(
```
brandonkachen: i think we have functions like this in our stripe helpers/webhooks, so we don't need to duplicate it

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,690 @@
...
+  const weekEnd = getWeekEnd(billingPeriodStart, now)
+
+  const result = await conn
+    .select({
+      total: sql<number>`COALESCE(SUM(${schema.creditLedger.principal} - ${schema.creditLedger.balance}), 0)`,
+    })
+    .from(schema.creditLedger)
+    .where(
+      and(
+        eq(
```
brandonkachen: users can "subscribe-churn" hack by canceling an existing subscription, renewing, then getting a different ID. we should instead filter on the `user_id` field

packages/internal/src/db/schema.ts
```diff
@@ -52,6 +52,12 @@ export const agentStepStatus = pgEnum('agent_step_status', [
   'skipped',
 ])
 
+export const subscriptionStatusEnum = pgEnum('subscription_status', [
```
brandonkachen, line 55: add these enums, just in case:

```
'incomplete'
'incomplete_expired'
'trialing'
'active'
'past_due'
'canceled'
'unpaid'
'paused'
```

packages/internal/src/db/schema.ts
```diff
@@ -77,6 +83,7 @@ export const user = pgTable('user', {
   auto_topup_threshold: integer('auto_topup_threshold'),
   auto_topup_amount: integer('auto_topup_amount'),
   banned: boolean('banned').notNull().default(false),
+  subscription_count: integer('subscription_count').notNull().default(0),
```
brandonkachen: what is the point of this? o.o feels very unnecessary, since we can just query over the `subscription` table

packages/internal/src/env-schema.ts
```diff
@@ -61,6 +62,7 @@ export const serverProcessEnv: ServerInput = {
   STRIPE_WEBHOOK_SECRET_KEY: process.env.STRIPE_WEBHOOK_SECRET_KEY,
   STRIPE_USAGE_PRICE_ID: process.env.STRIPE_USAGE_PRICE_ID,
   STRIPE_TEAM_FEE_PRICE_ID: process.env.STRIPE_TEAM_FEE_PRICE_ID,
+  STRIPE_SUBSCRIPTION_200_PRICE_ID: process.env.STRIPE_SUBSCRIPTION_200_PRICE_ID,
```
brandonkachen, line 68: make sure to update `.env.example`

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,690 @@
...
+  const now = new Date()
+
+  // Find all free/referral grants with remaining balance
+  const unusedGrants = await tx
+    .select()
+    .from(schema.creditLedger)
+    .where(
+      and(
+        eq(schema.creditLedger.user_id, userId),
+        inArray(schema.creditLedger.type, ['free', 'referral']),
```
brandonkachen: should remove this, otherwise it could artificially filter out rows we want. the `expires_at` check will ensure only expiring credits are migrated over (which is what we want).

web/src/app/api/stripe/webhook/route.ts
```diff
@@ -2,6 +2,10 @@ import {
   grantOrganizationCredits,
```
brandonkachen, line 2: this whole file feels messy; seems like codebuff is trying to shoehorn its conception of the code into the existing thing. we should have `codex` review and see if there's something better.

web/src/app/profile/components/usage-display.tsx
```diff
@@ -53,6 +53,14 @@ const grantTypeInfo: Record<
     label: 'Monthly Free',
     description: 'Your monthly allowance',
   },
+  subscription: {
+    bg: 'bg-indigo-500',
+    text: 'text-indigo-600 dark:text-indigo-400',
+    gradient: 'from-indigo-500/70 to-indigo-600/70',
+    icon: <Star className="h-4 w-4" />,
+    label: 'Pro Subscription',
+    description: 'Credits from your Pro plan',
```
brandonkachen: also make sure to rename this when we finalize on the name

brandonkachen:
oh and please make sure to run the db migration locally once you're happy with the db schema, so we have all the artifact files that drizzle will apply on prod when you merge

packages/billing/src/subscription.ts
jahooma: The query also currently includes `null` `expires_at rows`. Maybe I don't understand, but we only want to get the rows with an expires_at in the future?

common/src/constants/subscription-plans.ts
```diff
@@ -0,0 +1,33 @@
...
+    weeklyCreditsLimit: 4000,
+  },
+  200: {
+    monthlyPrice: 200,
+    creditsPerBlock: 1250,
+    blockDurationHours: 5,
+    weeklyCreditsLimit: 12500,
+  },
+  500: {
+    monthlyPrice: 500,
```
brandonkachen, line 24: shouldn't we be more generous with higher paying users? that seems to be standard in the industry

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * Maps a Stripe subscription status to our local enum.
+ */
+function mapStripeStatus(status: Stripe.Subscription.Status): SubscriptionStatus {
+  const validStatuses: readonly string[] = schema.subscriptionStatusEnum.enumValues
+  if (validStatuses.includes(status)) return status as SubscriptionStatus
+  return 'incomplete'
+}
+
+const priceToTier: Record<string, SubscriptionTierPrice> = {
```
brandonkachen: this ought to be derived from `SUBSCRIPTION_TIERS` in common/src/constants/subscription-plans.ts, no? 

also, there shouldn't be a need to check truthiness for these env vars, since they're required by zod to be strings of length 1+

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+
+function getTierFromPriceId(priceId: string): SubscriptionTierPrice | null {
+  return priceToTier[priceId] ?? null
+}
+
+const tierToPrice = Object.fromEntries(
+  Object.entries(priceToTier).map(([priceId, tier]) => [tier, priceId]),
+) as Partial<Record<SubscriptionTierPrice, string>>
+
+export function getTierPriceId(tier: SubscriptionTierPrice): string | null {
```
brandonkachen: dupe of `getTierFromPriceId`

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+
+// ---------------------------------------------------------------------------
+// invoice.paid
+// ---------------------------------------------------------------------------
+
+/**
+ * Handles a paid invoice for a subscription.
+ *
+ * - On first payment (`subscription_create`): calls `handleSubscribe` to
+ *   migrate the user's renewal date and unused credits (Option B).
```
brandonkachen: what is option b?

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+  if (!priceId) {
+    logger.error(
+      { subscriptionId },
+      'Stripe subscription has no price on first item',
+    )
+    return
+  }
+
+  // Look up the user for this customer
+  const userId = (await getUserByStripeCustomerId(customerId))?.id ?? null
```
brandonkachen: nit: i hate this pattern of wrapping awaits in parentheses, using a const for the awaited value is cleaner

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+      status: 'active',
+      billing_period_start: new Date(stripeSub.current_period_start * 1000),
+      billing_period_end: new Date(stripeSub.current_period_end * 1000),
+      cancel_at_period_end: stripeSub.cancel_at_period_end,
+    })
+    .onConflictDoUpdate({
+      target: schema.subscription.stripe_subscription_id,
+      set: {
+        status: 'active',
+        ...(userId ? { user_id: userId } : {}),
```
brandonkachen: if userId isn't defined, something really wrong has happened, and we should probably error out before getting here.

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+export async function handleSubscriptionInvoicePaymentFailed(params: {
+  invoice: Stripe.Invoice
+  logger: Logger
+}): Promise<void> {
+  const { invoice, logger } = params
+
+  if (!invoice.subscription) return
+  const subscriptionId = getStripeId(invoice.subscription)
+  const customerId = getStripeId(invoice.customer)
+  const userId = customerId
```
brandonkachen: nit: make a `let` initialized to null, then check `if (customerId)` and set userId if so

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+    .update(schema.subscription)
+    .set({
+      status: 'past_due',
+      updated_at: new Date(),
+    })
+    .where(eq(schema.subscription.stripe_subscription_id, subscriptionId))
+
+  trackEvent({
+    event: AnalyticsEvent.SUBSCRIPTION_PAYMENT_FAILED,
+    userId: userId ?? 'system',
```
brandonkachen, line 206: "system" is ambiguous, we can say "non-existent user" or something

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+  if (!priceId) {
+    logger.error(
+      { subscriptionId },
+      'Subscription update has no price — skipping',
+    )
+    return
+  }
+
+  const customerId = getStripeId(stripeSubscription.customer)
+  const userId = (await getUserByStripeCustomerId(customerId))?.id ?? null
```
brandonkachen: if userId is null, we should probably error

packages/billing/src/subscription-webhooks.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  trackEvent({
+    event: AnalyticsEvent.SUBSCRIPTION_CANCELED,
+    userId: userId ?? 'system',
+    properties: { subscriptionId },
+    logger,
+  })
+
+  logger.info({ subscriptionId }, 'Subscription canceled')
+}
```
brandonkachen, line 377: these handlers all do the same thing; get stripe user info, update db, then track event. it _might_ be cleaner to have one function that does the high level stuff and then branches out depending on the update type. 

but iono, maybe we just need to pass in the actual data we care about into these helper functions instead

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
+import { trackEvent } from '@codebuff/common/analytics'
```
brandonkachen, line 1: we should def add a lot of unit tests for these functions

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
...
+
+  const totalUnused = unusedGrants.reduce(
+    (sum, grant) => sum + grant.balance,
+    0,
+  )
+
+  // Deterministic ID ensures idempotency — duplicate webhook deliveries
+  // will hit onConflictDoNothing and the handleSubscribe caller checks
+  // for this operation_id before running.
+  const operationId = `subscribe-migrate-${subscriptionId}`
```
brandonkachen, line 707: just thinking out loud: if the user cancels their subscription and then creates a new one, they'll get a different operationId. which is fine because we'll just migrate their pre-existing credits, we're not granting/removing any.

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
...
+
+  if (totalUnused === 0) {
```
brandonkachen, line 709: this doesn't feel necessary at all; zero unused fits into the general case of migrating credits right?

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
...
+    logger.debug({ userId }, 'No unused credits to migrate')
+    return
+  }
+
+  // Zero out old grants
+  for (const grant of unusedGrants) {
+    await tx
+      .update(schema.creditLedger)
+      .set({ balance: 0 })
+      .where(eq(schema.creditLedger.operation_id, grant.operation_id))
```
brandonkachen, line 737: we should update the description to say how much we're migrating, for future reference

packages/internal/src/db/schema.ts
```diff
@@ -442,6 +459,59 @@ export const adImpression = pgTable(
...
 
+// Subscription tables
+export const subscription = pgTable(
+  'subscription',
+  {
+    stripe_subscription_id: text('stripe_subscription_id').primaryKey(),
+    stripe_customer_id: text('stripe_customer_id').notNull(),
+    user_id: text('user_id').references(() => user.id, { onDelete: 'cascade' }),
+    stripe_price_id: text('stripe_price_id').notNull(),
+    tier: integer('tier'),
```
brandonkachen, line 470: is the idea of "tier" to be a number corresponding to the dollar amount we charge users?

packages/internal/src/db/schema.ts
```diff
@@ -442,6 +459,59 @@ export const adImpression = pgTable(
...
+    status: subscriptionStatusEnum('status').notNull().default('active'),
```
brandonkachen, line 472: since this can be cancelled or paused or something else, having a `canceled_at` field feels slightly like we'll be allowing for bad states to happen (e.g. status = `canceled` but canceled_at = null). dunno how we'd solve this though. and in practice it's not likely to be a big issue

packages/internal/src/db/schema.ts
```diff
@@ -442,6 +459,59 @@ export const adImpression = pgTable(
...
+      .where(sql`${table.status} = 'active'`),
+  ],
+)
+
+export const limitOverride = pgTable('limit_override', {
+  user_id: text('user_id')
+    .primaryKey()
+    .references(() => user.id, { onDelete: 'cascade' }),
+  credits_per_block: integer('credits_per_block').notNull(),
+  block_duration_hours: integer('block_duration_hours').notNull(),
```
brandonkachen, line 506: is integer ok, or are we likely to change this to a decimal later?

packages/internal/src/db/schema.ts
```diff
@@ -442,6 +459,59 @@ export const adImpression = pgTable(
...
+  weekly_credit_limit: integer('weekly_credit_limit').notNull(),
```
brandonkachen, line 507: nit: would be nice for this to mirror the name `credits_per_block` more, e.g. `credits_per_week`

packages/internal/src/util/stripe.ts
```diff
@@ -1,6 +1,18 @@
...
+import { eq } from 'drizzle-orm'
 import Stripe from 'stripe'
 
-import { env } from '@codebuff/internal/env'
+/**
+ * Extracts the ID string from a Stripe expandable field.
+ */
+export function getStripeId(expandable: string | { id: string }): string
+export function getStripeId(expandable: string | { id: string } | null | undefined): string | undefined
+export function getStripeId(expandable: string | { id: string } | null | undefined): string | undefined {
```
brandonkachen: is method overloading necessary here?

web/src/app/api/stripe/cancel-subscription/route.ts
```diff
@@ -0,0 +1,60 @@
...
+      )
+
+    logger.info(
+      { userId, subscriptionId: subscription.stripe_subscription_id },
+      'Subscription set to cancel at period end',
+    )
+
+    return NextResponse.json({ success: true })
+  } catch (error: unknown) {
+    const message =
```
brandonkachen: this error handling feels sus, i wonder if we should split the try catch between the two `await`s and handle them differently

web/src/app/api/stripe/change-subscription-tier/route.ts
```diff
@@ -0,0 +1,161 @@
...
+      { error: 'Your account has been suspended. Please contact support.' },
+      { status: 403 },
+    )
+  }
+
+  const body = await req.json().catch(() => null)
+  const rawTier = Number(body?.tier)
+  if (!rawTier || !(rawTier in SUBSCRIPTION_TIERS)) {
+    return NextResponse.json(
+      { error: 'Invalid tier. Must be 100, 200, or 500.' },
```
brandonkachen: we should not hardcode these values, but instead read from `SUBSCRIPTION_TIERS`

web/src/app/api/stripe/create-subscription/route.ts
```diff
@@ -0,0 +1,111 @@
...
+    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 })
+  }
+
+  const userId = session.user.id
+
+  const body = await req.json().catch(() => ({}))
+  const rawTier = Number(body.tier)
+  const tier = (rawTier && rawTier in SUBSCRIPTION_TIERS
+    ? rawTier
+    : 200) as SubscriptionTierPrice
```
brandonkachen: wait are we defaulting users to the most expensive tier if we don't find the tier they intended to create?

packages/billing/src/__tests__/usage-service.test.ts
```diff
@@ -49,6 +49,10 @@ describe('usage-service', () => {
           }),
         }))
 
+        await mockModule('@codebuff/billing/subscription', () => ({
```
brandonkachen, line 52: aren't we trying to avoid mock modules?

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
...
+import { AnalyticsEvent } from '@codebuff/common/constants/analytics-events'
+import { GRANT_PRIORITIES } from '@codebuff/common/constants/grant-priorities'
+import {
+  DEFAULT_TIER,
+  SUBSCRIPTION_DISPLAY_NAME,
+  SUBSCRIPTION_TIERS,
+} from '@codebuff/common/constants/subscription-plans'
+
+import type { SubscriptionTierPrice } from '@codebuff/common/constants/subscription-plans'
+import db from '@codebuff/internal/db'
```
brandonkachen, line 11: we should do dependency injection so we can more easily test

packages/billing/src/subscription.ts
```diff
@@ -0,0 +1,752 @@
...
+        gt(schema.creditLedger.expires_at, now),
+        gt(schema.creditLedger.balance, 0),
+      ),
+    )
+    .returning({ operation_id: schema.creditLedger.operation_id })
+
+  if (expired.length > 0) {
+    logger.info(
+      { userId, subscriptionId, expiredCount: expired.length },
+      'Expired active block grants for tier change',
```
brandonkachen: technically this function doesn't know it's for a tier change, so we shouldn't log the last part

web/src/app/api/stripe/change-subscription-tier/route.ts
```diff
@@ -0,0 +1,161 @@
...
+        db
+          .update(schema.subscription)
+          .set({ tier, stripe_price_id: newPriceId, updated_at: new Date() })
+          .where(
+            eq(
+              schema.subscription.stripe_subscription_id,
+              subscription.stripe_subscription_id,
+            ),
+          ),
+        expireActiveBlockGrants({
```
brandonkachen: not necessarily an issue, but just for my comprehension: now when a user changes their subscription tier, they'll get the full block size. so if they used up an entire block and upgrade, they'll immediately get a new, full block at the new tier's limits.

in theory, a user could upgrade and use a block in a short time, then downgrade and use another block, then repeat. and while they'll still be charged for the time spent on each tier, they will be able to circumvent the block limits. 

in practice, it's unlikely users will do something like this and we'd still have the weekly limit to catch them.

web/src/app/api/stripe/create-subscription/route.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  try {
+    const checkoutSession = await stripeServer.checkout.sessions.create({
+      customer: user.stripe_customer_id,
+      mode: 'subscription',
+      line_items: [{ price: priceId, quantity: 1 }],
+      allow_promotion_codes: true,
+      success_url: `${env.NEXT_PUBLIC_CODEBUFF_APP_URL}/profile?tab=usage&subscription_success=true`,
+      cancel_url: `${env.NEXT_PUBLIC_CODEBUFF_APP_URL}/strong?canceled=true`,
```
brandonkachen, line 78: i don't think we have a webpage for this?

web/src/app/api/stripe/create-subscription/route.ts
```diff
@@ -0,0 +1,111 @@
...
+    )
+  }
+
+  try {
+    const checkoutSession = await stripeServer.checkout.sessions.create({
+      customer: user.stripe_customer_id,
+      mode: 'subscription',
+      line_items: [{ price: priceId, quantity: 1 }],
+      allow_promotion_codes: true,
+      success_url: `${env.NEXT_PUBLIC_CODEBUFF_APP_URL}/profile?tab=usage&subscription_success=true`,
```
brandonkachen, line 77: we should def make sure to set up that page to check for `subscription_success=true` cases on this page and show some welcome/thanks for subscribing messages

web/src/app/api/stripe/create-subscription/route.ts
```diff
@@ -0,0 +1,111 @@
...
+      cancel_url: `${env.NEXT_PUBLIC_CODEBUFF_APP_URL}/strong?canceled=true`,
+      metadata: {
+        userId,
+        type: 'strong_subscription',
+      },
+      subscription_data: {
+        description: `Codebuff Strong — $${tier}/mo unlimited coding sessions`,
```
brandonkachen: "unlimited coding sessions" sounds weird, tbh. "unlimited code-gen" sounds better

web/src/app/api/stripe/webhook/route.ts
```diff
@@ -354,17 +372,27 @@ const webhookHandler = async (req: NextRequest): Promise<NextResponse> => {
       case 'customer.created':
         break
       case 'customer.subscription.created':
-      case 'customer.subscription.updated':
+      case 'customer.subscription.updated': {
+        const sub = event.data.object as Stripe.Subscription
+        if (sub.metadata?.organization_id) {
```
brandonkachen, line 377: aren't we deprecating the orgs stuff? if so, we can just error out earlier on in the webhook, instead of running code we don't want to maintain and cluttering up each of these cases

brandonkachen (approved):
looks good! most of my comments are code readability nits/suggestions, so that means we're on the right track!

common/src/constants/subscription-plans.ts
jahooma: I updated so it is the most generous tier. But the credits per dollar accelerates most for the 200 plan, which I think most will want.

(If we increase any more for $500, we lose too much. If we lower the 200 plan, then users will be unsatisfied.)

packages/billing/src/subscription-webhooks.ts
jahooma: Actually it's the reverse operation. But I renamed to getPriceIdFromTier for clarity

packages/billing/src/subscription-webhooks.ts
jahooma: Good suggestion

packages/billing/src/subscription-webhooks.ts
jahooma: Removing references to it.

packages/billing/src/subscription-webhooks.ts
jahooma: Done

packages/billing/src/subscription-webhooks.ts
jahooma: Hmm, I see that, but  2/3 of the code is different, and I think this might be simpler/clearer

packages/internal/src/db/schema.ts
jahooma: Yup!

web/src/app/api/stripe/change-subscription-tier/route.ts
```diff
@@ -0,0 +1,161 @@
...
+        { error: 'Subscription configuration error.' },
+        { status: 500 },
+      )
+    }
+
+    await stripeServer.subscriptions.update(
+      subscription.stripe_subscription_id,
+      {
+        items: [{ id: itemId, price: newPriceId }],
+        proration_behavior: 'create_prorations',
```
brandonkachen: IMPORTANT: this is incorrect. We want upgrades prorate now, downgrades scheduled at period end. ChatGPT says:

**Upgrades**
Use:
- proration_behavior: "always_invoice"
- (recommended) payment_behavior: "pending_if_incomplete" to avoid granting the upgrade if payment fails

**Downgrades**
Don’t “change price now”. Instead, schedule the downgrade for current_period_end using a Subscription Schedule. Stripe explicitly suggests subscription schedules if you’re changing a subscription at the end of the billing period.

Conceptually:
- Phase 1: current price until current_period_end
- Phase 2: cheaper price starting at current_period_end
Set the phase transition proration_behavior: "none" so there’s no proration credit at the transition

This gives you exactly: “stay on current tier until renewal, then downgrade, no credit”.

web/src/app/api/stripe/change-subscription-tier/route.ts
```diff
@@ -0,0 +1,161 @@
...
+      },
+    )
+
+    try {
```
brandonkachen, line 121: maybe this try/catch block can go into a webhook; this function/route can just focus on being a wrapper around stripe, and then the webhook can deal with the consequences.

this would let us organize our logic using:
- routes as proxies for stripe actions 
- webhooks to propagate changes from stripe to our system

web/src/app/api/stripe/change-subscription-tier/route.ts
brandonkachen: link: https://chatgpt.com/c/697aa040-ea04-8321-823d-baf374188e7c#:~:text=Upgrades%20now%2C%20Downgrades%20scheduled%20at%20period%20end

web/src/app/api/stripe/change-subscription-tier/route.ts
jahooma: Yup, good to think about. Codebuff also thinks there's no issue b/c upgrading bills you immediately for the rest of the term, and the weekly limit is a hard cap.

web/src/app/api/stripe/create-subscription/route.ts
jahooma: Good catch

web/src/app/api/stripe/create-subscription/route.ts
jahooma: I created one but removed it from this PR. It's in the next one: https://github.com/CodebuffAI/codebuff/pull/424

web/src/app/api/stripe/create-subscription/route.ts
jahooma: Yep, added a note to the next PR

packages/billing/src/subscription.ts
jahooma: It's so that if this migration is tried a second time, it will fail because the operation id has already been used to save an empty credits block. It's not a terrible idea I think

packages/internal/src/db/schema.ts
jahooma: Good question, but I think integer is fine

packages/internal/src/db/schema.ts
jahooma: Reasonable, but weekly_credit_limit is clear enough

web/src/app/api/stripe/cancel-subscription/route.ts
jahooma: Done

packages/billing/src/subscription.ts
jahooma: Done

packages/billing/src/subscription.ts
jahooma: Done

packages/billing/src/subscription.ts
jahooma: Done

web/src/app/api/stripe/change-subscription-tier/route.ts
jahooma: Done

packages/billing/src/__tests__/usage-service.test.ts
jahooma: Maybe, but it is using our own mock wrapper which we are currently resetting between tests, and other tests in this file use it. So seems fine

packages/internal/src/db/schema.ts
jahooma: Yeah, I think that would be just be a logic bug we'd have to fix. I think having the status in the db is helpful overall

PR #424, opened 2026-01-28
Subscription client changes

jahooma:


cli/src/components/bottom-status-line.tsx
```diff
@@ -12,70 +13,141 @@ interface BottomStatusLineProps {
...
       : theme.muted
 
+  // Subscription remaining percentage (based on weekly)
+  const subscriptionRemaining = subscriptionRateLimit
+    ? 100 - subscriptionRateLimit.weeklyPercentUsed
+    : null
+  const isSubscriptionLimited = subscriptionRateLimit?.limited === true
+
+  // Get subscription reset time
+  const subscriptionResetTime = subscriptionRateLimit
```
brandonkachen: not a fan of all these nested ternaries, would be worthwhile (imo) to ask codebuff to replace them all with reasonably named constants and if blocks or, more ideally, states that we care about and reason about more clearly

cli/src/components/bottom-status-line.tsx
brandonkachen, line 1: just a personal preference thing, but i really don't care for bottom status bars. they take up a lot of vertical space (esp right now, since ours only shows subscriptions) and the info they show can still make users feel like they have to ration usage.

i don't mind that claude code/codex/kimi code hide this info in the cli and you have to go to the website to see it

cli/src/components/bottom-status-line.tsx
```diff
@@ -12,70 +13,141 @@ interface BottomStatusLineProps {
...
+        >
+          <text style={{ fg: theme.error }}>●</text>
+          <text style={{ fg: theme.muted }}> Claude</text>
+          {claudeResetTime && (
+            <text style={{ fg: theme.muted }}>{` · resets in ${formatResetTime(claudeResetTime)}`}</text>
+          )}
+        </box>
+      )}
+
+      {/* Show Codebuff Strong when subscribed and Claude not healthy */}
```
brandonkachen: shouldn't we display both at the same time? if a user has both subs active, they may want to see how both are being used (esp if we use more non-claude models in the future). again, i'm still against the need for this component in the first place

cli/src/components/subscription-limit-banner.tsx
```diff
@@ -0,0 +1,190 @@
...
+              </text>
+            )}
+          </>
+        ) : isBlockExhausted ? (
+          <>
+            <text style={{ fg: theme.warning }}>
+              ⏱️  Block limit reached
+            </text>
+            <text style={{ fg: theme.muted }}>
+              You've used all {rateLimit.blockLimit?.toLocaleString()} credits in this 5-hour block.
```
brandonkachen: nit: we should use the constant for the block size.

brandonkachen: also, if `rateLimit.blockLimit` doesn't exist, we should put a placeholder string, e.g. "your" -> "You've used all <your> credits..."

cli/src/components/subscription-limit-banner.tsx
```diff
@@ -0,0 +1,190 @@
...
+            </text>
+          </Button>
+        )}
+
+        <box style={{ flexDirection: 'row', gap: 2, marginTop: 1 }}>
+          {hasAlaCarteCredits ? (
+            <>
+              <Button onClick={handleContinueWithCredits}>
+                <text style={{ fg: theme.foreground }}>
+                  [Continue with a-la-carte] ({remainingBalance.toLocaleString()} available)
```
brandonkachen: should we make these look like proper buttons? like our `/feedback` submit button?

cli/src/components/subscription-limit-banner.tsx
```diff
@@ -0,0 +1,190 @@
...
+              )}
+            </>
+          ) : (
+            <>
+              <text style={{ fg: theme.muted }}>No a-la-carte credits available.</text>
+              <Button onClick={handleBuyCredits}>
+                <text style={{ fg: theme.foreground }}>[Buy Credits]</text>
+              </Button>
+              <Button onClick={handleWait}>
+                <text style={{ fg: theme.muted }}>[Wait]</text>
```
brandonkachen: "wait" feels like a weird thing to say by itself. and since the user can really only do one thing, why don't we just get rid of this button and add an `x` at the top right to close out the modal?

cli/src/components/usage-banner.tsx
```diff
@@ -99,12 +105,68 @@ export const UsageBanner = ({ showTime }: { showTime: number }) => {
   const adCredits = activeData.balanceBreakdown?.ad
   const renewalDate = activeData.next_quota_reset ? formatRenewalDate(activeData.next_quota_reset) : null
 
+  const hasSubscription = subscriptionData?.hasSubscription === true
+  const rateLimit = subscriptionData?.rateLimit
+  const subscriptionInfo = subscriptionData?.subscription
```
brandonkachen: code smell: if `hasSubscription` is true, we should have `subscriptionInfo`, but that may not be true here. we should derive the former from the latter imo

cli/src/components/usage-banner.tsx
```diff
@@ -99,12 +105,68 @@ export const UsageBanner = ({ showTime }: { showTime: number }) => {
...
+
   return (
     <BottomBanner
       borderColorKey={isLoadingData ? 'muted' : colorLevel}
       onClose={() => setInputMode('default')}
     >
       <box style={{ flexDirection: 'column', gap: 0 }}>
+        {/* Strong subscription section - only show if subscribed */}
```
brandonkachen: this whole section has too many nested ternaries; we can split them out into different cases and return different components from them.

cli/src/data/slash-commands.ts
```diff
@@ -73,6 +73,12 @@ export const SLASH_COMMANDS: SlashCommand[] = [
     description: 'View credits and subscription quota',
```
brandonkachen, line 73: unrelated to this change, but it's weird that we have a separate command registry from our slash commands; feels like a lot of unnecessary code duplication

cli/src/utils/block-operations.ts
```diff
@@ -12,6 +12,17 @@ import type {
   TextContentBlock,
 } from '../types/chat'
 
+const SHORT_THINKING_LINE_THRESHOLD = 5
```
brandonkachen: seems like your thinking ux changes snuck into this PR!

cli/src/utils/settings.ts
```diff
@@ -20,6 +20,7 @@ const DEFAULT_SETTINGS: Settings = {
 export interface Settings {
   mode?: AgentMode
   adsEnabled?: boolean
+  alwaysUseALaCarte?: boolean
```
brandonkachen, line 24: how do users set this value? do they run a slash command, or toggle in the web dashboard?

web/src/app/api/v1/chat/completions/_post.ts
```diff
@@ -182,6 +191,22 @@ export async function postChatCompletions(params: {
       logger,
     })
 
+    // For subscribers, ensure a block grant exists before checking balance.
+    // This is done here block grants should only start when the user begins working.
+    if (getActiveSubscription && ensureActiveBlockGrant) {
+      try {
```
brandonkachen: we could combine these two checks into one, or skip the activeSub check completely

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+    weeklyPercentUsed: number
+  }
+  limits?: {
+    creditsPerBlock: number
+    blockDurationHours: number
+    weeklyCreditsLimit: number
+  }
+}
+
+function formatDaysHours(dateStr: string): string {
```
brandonkachen: i feel like we have this function (or something very similar) defined elsewhere already.

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+  data,
+  email,
+}: {
+  data: SubscriptionApiResponse
+  email: string
+}) {
+  const { subscription, rateLimit } = data
+
+  const isCanceling = subscription?.cancelAtPeriodEnd
+  const billingPortalUrl = `${env.NEXT_PUBLIC_STRIPE_CUSTOMER_PORTAL}?prefilled_email=${encodeURIComponent(email)}`
```
brandonkachen: would be nice to send users to a portal that they're already logged into! i think we have to call the stripe sdk to get it though. then we could use `NEXT_PUBLIC_STRIPE_CUSTOMER_PORTAL` as a backup

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  return (
+    <Card className="max-w-xl">
+      <CardHeader className="pb-5">
+        <div className="flex items-center justify-between">
+          <CardTitle className="flex items-baseline gap-2 text-lg">
+            <span>💪</span>
+            {SUBSCRIPTION_DISPLAY_NAME}
+            <span className="text-sm font-normal text-muted-foreground">
+              ${subscription?.tier ?? 200}/mo
```
brandonkachen: not a fan of a hardcoded, default value. if `subscription` is missing, we should just show an error component

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+                </span>
+                <span className="text-muted-foreground">
+                  {rateLimit.blockLimit != null && rateLimit.blockUsed != null && rateLimit.blockLimit > 0
+                    ? `${Math.round(100 - (rateLimit.blockUsed / rateLimit.blockLimit) * 100)}%`
+                    : '100%'}
+                </span>
+              </div>
+              <ProgressBar
+                percentAvailable={
+                  rateLimit.blockLimit != null && rateLimit.blockUsed != null && rateLimit.blockLimit > 0
```
brandonkachen: if this evaluates to false, we'll show 100, which feels wrong

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+                    ? 100 - (rateLimit.blockUsed / rateLimit.blockLimit) * 100
```
brandonkachen: wait we shouldn't pass back the actual amounts from the server to the client; only the server should know how much is actually used, the client should only ever see percentages

web/src/app/profile/components/subscription-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+        <div className="flex items-start gap-3">
+          <div className="mt-0.5 rounded-lg bg-indigo-100 p-2 dark:bg-indigo-900/30">
+            <span className="text-xl">💪</span>
+          </div>
+          <div>
+            <h3 className="font-semibold">
+              Upgrade to {SUBSCRIPTION_DISPLAY_NAME}
+            </h3>
+            <p className="text-sm text-muted-foreground">
+              From $100/mo · Work in focused 5-hour sessions with no
```
brandonkachen: we're not releasing tiers yet, right? so this number should be the tier we actually want to show users. and if we're gonna hardcode something, it should be the $200/mo plan.

web/src/app/strong/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import StrongClient from './strong-client'
+
+import type { Metadata } from 'next'
+
+export async function generateMetadata(): Promise<Metadata> {
+  const canonicalUrl = `${env.NEXT_PUBLIC_CODEBUFF_APP_URL}/strong`
+  const title = `Codebuff ${SUBSCRIPTION_DISPLAY_NAME} — The Strongest Coding Agent`
+  const description =
+    'Deep thinking, multi-agent orchestration, and the strongest coding agent. Plans from $100/mo.'
```
brandonkachen: hardcoded `$100` again

packages/billing/src/subscription-webhooks.ts
```diff
@@ -375,3 +373,173 @@ export async function handleSubscriptionDeleted(params: {
...
+    )
+    return
+  }
+
+  // Update the subscription with the scheduled tier
+  const result = await db
+    .update(schema.subscription)
+    .set({
+      scheduled_tier: scheduledTier,
+      updated_at: new Date(),
```
brandonkachen: it would be nice to build `updated_at` on the postgres layer, instead of having to set it each time like this. i think we'd have to create a new migration and write sql code, but that's how it works for most/all other `updated_at` fields 🤷‍♂️

packages/billing/src/subscription-webhooks.ts
```diff
@@ -375,3 +373,173 @@ export async function handleSubscriptionDeleted(params: {
...
+  if (!subscriptionId) {
+    logger.warn(
+      { scheduleId: schedule.id },
+      'Subscription schedule has no linked subscription — skipping',
+    )
+    return
+  }
+
+  // Need at least 2 phases to have a scheduled change (current + future)
+  if (!schedule.phases || schedule.phases.length < 2) {
```
brandonkachen, line 437: we should definitely test to make sure we always get 2+ phases. i'm not totally sure what that is

web/src/app/profile/components/subscription-section.tsx
jahooma: I don't think this matters that much

web/src/app/profile/components/subscription-section.tsx
jahooma: I was going to release with the tiers!

cli/src/components/bottom-status-line.tsx
jahooma: Removed bottom bar. Now shows when the subscription is used to pay in place of credits used at end of response

cli/src/components/usage-banner.tsx
jahooma: I don't think ui code matters much as long as it works

cli/src/data/slash-commands.ts
jahooma: Agreed, I think I remember only one being the "real" one lol

web/src/app/api/v1/chat/completions/_post.ts
jahooma: It gets the sub data, which is passed to the ensure active block grant. 

But I think we can clean this up by injecting a single function that does both of these

packages/billing/src/subscription-webhooks.ts
jahooma: Added docs on phases. I tested the downgrade/upgrade flows manually

cli/src/utils/settings.ts
jahooma: Added a way in /usage banner

web/src/app/profile/components/subscription-section.tsx
jahooma: Extracted util function

web/src/app/profile/components/subscription-section.tsx
jahooma: Cool, i added an authenticated link!

jahooma (commented):
Addressed comments!

PR #429, opened 2026-02-08
fix anthropic tool call bug

pranav100000:


packages/agent-runtime/src/tools/stream-parser.ts
```diff
@@ -84,6 +84,10 @@ export async function processStream(
     userId,
   } = params
   const fullResponseChunks: string[] = [fullResponse]
+  const messageHistoryBeforeStream = expireMessages(
+    agentState.messageHistory,
+    'agentStep',
+  )
```
jahooma: I don't think this is necessary, this is run right before each step already (in runAgentStep)

packages/agent-runtime/src/tools/stream-parser.ts
```diff
@@ -182,7 +188,7 @@ export async function processStream(
               : (toolName as ToolName),
             input: transformed ? transformed.input : input,
             fromHandleSteps: false,
-            skipDirectResultPush: isXmlMode,
+            skipDirectResultPush: true,
```
jahooma: Seems we should refactor to delete this param entirely if always true

packages/agent-runtime/src/tools/stream-parser.ts
```diff
@@ -327,20 +333,20 @@ export async function processStream(
...
-  agentState.messageHistory.push(...errorMessages)
+  // === FINALIZATION ===
+  // Build message history from the pre-stream snapshot so tool_calls and
+  // tool_results are always appended in deterministic order.
+  agentState.messageHistory = buildArray<Message>([
+    ...messageHistoryBeforeStream,
+    ...assistantMessages,
+    ...toolResultsToAddAfterStream,
+    ...errorMessages,
+  ])
```
jahooma, line 322: Conceptually, I think it does make more sense to do all the mutating of message history here, in one place.

packages/agent-runtime/src/tools/tool-executor.ts
```diff
@@ -560,6 +564,10 @@ export async function executeCustomToolCall(
 
       toolResults.push(toolResult)
 
+      if (!excludeToolFromMessageHistory) {
+        toolResultsToAddAfterStream.push(toolResult)
+      }
+
       if (!excludeToolFromMessageHistory && !params.skipDirectResultPush) {
         agentState.messageHistory.push(toolResult)
       }
```
jahooma: In the new paradigm, we can delete this if and
`agentState.messageHistory.push(toolResult)`.

We might want to rename toolResultsToAddAfterStream to toolResultsToAddToMessageHistory, since all tool results are added after stream and these are the ones included in the message history in particular

jahooma (commented):
Thanks Pranav! This is a tricky bug and our code here is kinda bad haha.

Made some comments, but the thrust of the change seems good.

If you make these fixes, I'll test & merge it!

PR #477, opened 2026-03-21
Fix/issue 367 kimi k2 streaming

### LINKED ISSUE #367: The “kimi-k2” Model breaks off after use at the beginning
see my prev topic
https://github.com/CodebuffAI/codebuff/issues/365#issuecomment-3536199250

ayorindeadunse:
## Description

Fixes #367 - The "kimi-k2" Model breaks off after use at the beginning

### Problem
The kimi-k2 model was breaking off during streaming responses due to truncated JSON chunks. When JSON is sent at newline boundaries, it can arrive incomplete, causing parsing failures and stream termination.

### Solution
This PR adds defensive JSON parsing and error recovery mechanisms to gracefully handle truncated or malformed JSON responses from OpenRouter:

1. **`attemptJSONParse()` function** - Implements multi-stage JSON recovery:
   - Attempts to parse JSON as-is
   - Removes trailing newlines from truncated responses
   - Tries closing incomplete JSON structures (}, ]}, }]}, etc.)
   - Returns null only after all recovery attempts fail

2. **Enhanced error handling** - Improved logging and validation:
   - Better detection of truncated chunks vs actual errors
   - Graceful fallback when parsing fails (logs and continues)
   - Detailed debugging information for troubleshooting

3. **Import cleanup** - Reorganized imports for consistency and removed unused types

### Testing
- Build succeeds with no TypeScript/type errors
- All existing tests pass
- No regressions detected

### Related
Fixes #367

### DIFF
--- a/cli/src/components/multiline-input.tsx
+++ b/cli/src/components/multiline-input.tsx
@@ -275,7 +275,7 @@ export const MultilineInput = forwardRef<
     ? Math.max(
       0,
       lineInfo.lineStartCols.findLastIndex(
-        (lineStart) => lineStart <= cursorPosition,
+        (lineStart: number) => lineStart <= cursorPosition,
       ),
     )
     : 0
@@ -803,7 +803,7 @@ export const MultilineInput = forwardRef<
       // Fall back to logical line boundaries if visual info is unavailable
       const lineStarts = currentLineInfo?.lineStartCols ?? []
       const visualLineIndex = lineStarts.findLastIndex(
-        (start) => start <= cursorPosition,
+        (start: number) => start <= cursorPosition,
       )
       const visualLineStart = visualLineIndex >= 0
         ? lineStarts[visualLineIndex]
--- a/cli/src/components/multiline-input.tsx
+++ b/cli/src/components/multiline-input.tsx
@@ -89,16 +89,16 @@ const TAB_WIDTH = 4
  */
 function isPrintableCharacterKey(key: KeyEvent): boolean {
   const name = key.name
-  
+
   // No name = likely multi-byte input (Chinese, Japanese, Korean, etc.) - treat as printable
   if (!name) return true
-  
+
   // Single character name = regular ASCII printable (a, b, 1, $, etc.)
   if (name.length === 1) return true
-  
+
   // Special case: space key has name 'space' but is printable
   if (name === 'space') return true
-  
+
   // Multi-char name = special key (up, f1, backspace, etc.)
   return false
 }
@@ -122,8 +122,8 @@ function renderPositionToOriginal(text: string, renderPos: number): number {
 
 type KeyWithPreventDefault =
   | {
-      preventDefault?: () => void
-    }
+    preventDefault?: () => void
+  }
   | null
   | undefined
 
@@ -136,9 +136,9 @@ function isAltModifier(key: KeyEvent): boolean {
   const ESC = '\x1b'
   return Boolean(
     key.option ||
-      (key.sequence?.length === 2 &&
-        key.sequence[0] === ESC &&
-        key.sequence[1] !== '['),
+    (key.sequence?.length === 2 &&
+      key.sequence[0] === ESC &&
+      key.sequence[1] !== '['),
   )
 }
 
@@ -241,9 +241,9 @@ export const MultilineInput = forwardRef<
 
   const lineInfo = textRef.current
     ? (
-        (textRef.current satisfies TextRenderable as any)
-          .textBufferView as TextBufferView
-      ).lineInfo
+      (textRef.current satisfies TextRenderable as any)
+        .textBufferView as TextBufferView
+    ).lineInfo
     : null
 
   // Focus/blur scrollbox when focused prop changes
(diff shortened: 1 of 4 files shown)


web/src/llm-api/openrouter.ts
```diff
@@ -470,24 +513,62 @@ async function handleLine({
...
-    obj = JSON.parse(raw)
-  } catch (error) {
-    logger.warn(
-      { error: getErrorObject(error, { includeRawError: true }) },
-      'Received non-JSON OpenRouter response',
+  // Parse the string into an object with defensive recovery
+  let obj = attemptJSONParse(raw)
+  if (obj === null) {
+    // All parsing attempts failed - log with detailed context for debugging
+    logger.debug(
```
jahooma, line 520: Should we keep the logger.warn level rather than downgrading to debug?

cli/src/components/multiline-input.tsx
```diff
@@ -89,16 +89,16 @@ const TAB_WIDTH = 4
  */
 function isPrintableCharacterKey(key: KeyEvent): boolean {
   const name = key.name
-  
+
```
jahooma, line 92: Note: it would be better not to include many trivial formatting changes

cli/src/components/multiline-input.tsx
```diff
@@ -271,13 +271,13 @@ export const MultilineInput = forwardRef<
     [],
   )
 
-  const cursorRow = lineInfo
+  const cursorRow = lineInfo && lineInfo.lineStartCols
```
jahooma, line 274: Are you sure that lineInfo doesn't include lineStartCols sometimes?

cli/src/components/multiline-input.tsx
ayorindeadunse: I'll work on this. Thanks!

cli/src/components/multiline-input.tsx
ayorindeadunse: I thought so, but typescript flagged an error. I'll check this again.

web/src/llm-api/openrouter.ts
```diff
@@ -470,24 +513,62 @@ async function handleLine({
...
+      // Try to handle as error even if schema doesn't match perfectly
+      logger.warn(
+        {
+          error: getErrorObject(parsed.error, { includeRawError: true }),
+          rawObject: JSON.stringify(obj).slice(0, 500),
+        },
+        'OpenRouter response matches error pattern but schema validation failed',
+      )
+      // Continue processing as error response
+      return handleResponse({
```
jahooma: If it's error like, doesn't handleResponse already return { state }, e.g. see:

if ('error' in data || !data.usage) {
    // Stream not finished
    return { state }
  }

So this doesn't seem to do anything

hiSandog:
One concern with the new `attemptJSONParse()` recovery in `web/src/llm-api/openrouter.ts`: appending `}`, `]}`, etc. can turn a truncated delta frame into syntactically valid JSON, and `handleStreamChunk()` will append that delta to `state.responseText` / `state.reasoningText` immediately.

If the real complete SSE frame arrives on the next read, the same token span can be counted twice or with corrupted content. That feels riskier than just skipping the malformed fragment. A safer recovery path would be to keep an unfinished `data:` payload in a pending buffer and retry parsing on the next read, or to limit recovery to terminal usage/error frames where duplicate token emission cannot corrupt the accumulated state.

codebuff-team:
Thanks for digging into #367 — streaming breakage with kimi-k2 is a real and annoying bug. Unfortunately this PR can't be ported as-is for a few reasons:

1. **Out of scope**: the actual fix is in `web/src/llm-api/openrouter.ts`. This mirror explicitly excludes changes under `web/`, so this diff can't be reviewed/merged here regardless of correctness.

2. **Unrelated changes bundled in**: `package.json`/`bun.lock` add a `geoip-lite` dependency that has nothing to do with kimi-k2 streaming, and `cli/src/components/multiline-input.tsx` has a large block of pure reformatting (indentation/whitespace) plus an unexplained behavioral tweak to `cursorRow`/`lineInfo.lineStartCols` guards. None of this belongs in a PR titled "fix kimi-k2 streaming" — it makes the diff much harder to review and suggests a messy branch/rebase.

3. **The core approach is risky even on its own merits.** `attemptJSONParse()` guesses at closing truncated JSON by blindly appending `}`, `]}`, `}]}`, `[]` to the raw string. Appending brackets to a genuinely truncated fragment doesn't reliably reconstruct the original object — it can produce JSON that parses successfully but represents different data than what OpenRouter intended, and that gets silently fed into stream state. If the real problem is that JSON chunks arrive split across newline-delimited SSE boundaries, the correct fix is in the chunk-buffering/line-splitting logic that assembles fragments before parse, not in speculative JSON repair after the fact.

If you want to pursue this further, I'd suggest: (a) isolate the fix to just the streaming/parsing logic, no unrelated dependency or formatting changes, (b) find where lines are split on newlines and confirm whether partial JSON is being handed to `JSON.parse` before a full line is buffered — that's almost certainly the actual bug, and (c) add a regression test reproducing a truncated OpenRouter SSE chunk.

codebuff-team:
Marking this stale - there has been no activity here for 154 days. It will close in 7 days unless someone comments.

codebuff-team:
Closing this as stale - no activity since it was flagged 7 days ago. Reopen it any time if it is still relevant; that is not a hassle and we would rather have it open than lost.

PR #506, opened 2026-04-16
fix: avoid DNS lookup after proxied release CONNECT

YoungSx:
## Summary
- extract the release launcher HTTPS proxy logic into dedicated `http.js` helpers for `codebuff`, `codecane`, and `freebuff`
- switch proxied requests from a raw `createConnection` callback to a custom `https.Agent`, so requests reuse the CONNECT tunnel without falling back to local DNS lookup
- add a regression test covering all three release helpers and include the helper file in each published npm package

## Verification
- `bun --cwd cli test src/__tests__/release/proxy-http-get.test.ts`
- manual live check: patched `freebuff/cli/release/http.js` fetched `https://registry.npmjs.org/freebuff/latest` successfully through the configured proxy (status 200)

## Root cause
The release launchers established the proxy CONNECT tunnel correctly, but then passed a raw `createConnection` callback into `https.get`. In proxy environments like WSL, Node still attempted a fresh hostname lookup for `registry.npmjs.org`, causing first-run startup to fail with `EAI_AGAIN` even though the proxy tunnel was already open.

### DIFF
--- a/cli/release-staging/http.js
+++ b/cli/release-staging/http.js
@@ -1,0 +1,176 @@
+const http = require('http')
+const https = require('https')
+const tls = require('tls')
+
+function createReleaseHttpClient({
+  env = process.env,
+  userAgent,
+  requestTimeout,
+  httpModule = http,
+  httpsModule = https,
+  tlsModule = tls,
+}) {
+  function getProxyUrl() {
+    return (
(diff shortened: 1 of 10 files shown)


jahooma:
Thanks for the fix!

PR #508, opened 2026-04-18
fix: correct code-map line counting

hiSandog:
## Summary
- fix parseTokens() so multiline files count newlines correctly by applying the nullish fallback before adding 1
- update the code-map regression test to expect the correct line count for a 3-line input

## Validation
- reproduced the root cause locally with Node: the old expression returns 2 for a 3-line string, while the fixed expression returns 3
- full bun test packages/code-map/__tests__/parse.test.ts could not be run in this environment because bun is not installed locally


jahooma:
Awesome, thanks!

PR #571, opened 2026-04-30
Add Fireworks fallback for CanopyWave

jahooma:
Adds a server-side fallback from CanopyWave to Fireworks for supported models when CanopyWave returns availability failures such as 429, 5xx, or "No available workers".
If CanopyWave is not configured, supported models now route directly to Fireworks instead of failing on the missing key.
This keeps the existing provider billing/logging paths intact and covers both streaming and non-streaming completions.
Validated with `bun test web/src/app/api/v1/chat/completions/__tests__/completions.test.ts` and `bun run --cwd web typecheck`.


codebuff-team:
This PR modifies files under `web/src/app/api/v1/chat/completions/_post.ts` and its test file, which fall under the `web/` path excluded from this mirror's review scope per repo policy. Regardless of the change's technical merit (the CanopyWave→Fireworks fallback logic and the added test coverage for both streaming and non-streaming paths look reasonable and well-tested), this repo cannot accept changes to `web/`. If this is meant to be ported to the private tree, it should go through that internal process rather than this mirror.

PR #615, opened 2026-05-07
Add OpenCode runner to BuffBench

brandonkachen:
## Summary
- add an `external:opencode` BuffBench runner backed by the OpenCode CLI JSON stream
- wire OpenCode into external agent selection and runner exports
- document the OpenCode CLI prerequisites and default/override model behavior

## Validation
- `bun run --cwd evals typecheck`
- `bun run --cwd evals test`
- `bun run typecheck`

I also ran `bun run test`; it failed in unrelated CLI test suites with unhandled errors between tests, while the web suite passed when `.env.local` was loaded.

### DIFF
--- a/evals/buffbench/README.md
+++ b/evals/buffbench/README.md
@@ -139,6 +139,7 @@ BuffBench supports running external CLI coding agents for comparison:
 
 - **Claude Code**: Use `external:claude` - requires `claude` CLI installed
 - **Codex**: Use `external:codex` - requires `codex` CLI installed
+- **OpenCode**: Use `external:opencode` - requires `opencode` CLI installed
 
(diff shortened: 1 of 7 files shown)


brandonkachen:
Addressed reviewer feedback about OpenCode error events in `5320ea20c`.

The runner now reads `error.data.message` before falling back to `error.message` / `error.name`, so successful-exit OpenCode error events surface actionable messages like `Model not found: ...` instead of only `UnknownError`.

Validation:
- `bun run --cwd evals typecheck`
- `bun run --cwd evals test`
- `bun run typecheck`

Smoke check reproduced the invalid-model JSON event shape with `opencode run --format json --model opencode/definitely-not-a-model --agent build hi`.

PR #617, opened 2026-05-07
Restore state-backed tool call IDs

brandonkachen:
## Summary
- Restore deterministic tool call IDs using the `functions.<tool>.<index>` shape with a run/session-backed monotonic counter.
- Seed the counter from retained history and pending calls, including the older colon-delimited deterministic ID shape.
- Share the counter through spawned agent state while keeping the public agent template types free of internal allocator state.

## Validation
- `bun test packages/agent-runtime/src/util/__tests__/tool-call-id.test.ts packages/agent-runtime/src/__tests__/run-programmatic-step.test.ts packages/agent-runtime/src/__tests__/tool-validation-error.test.ts`
- `bun run --cwd packages/agent-runtime typecheck`
- `bun run --cwd packages/agent-runtime test`
- `NODE_ENV=production bun test cli/src/utils/__tests__/sdk-event-handlers.test.ts cli/src/utils/__tests__/message-block-helpers.test.ts cli/src/utils/__tests__/send-message-helpers.test.ts`
- `bun run --cwd cli src/index.tsx --help`


codebuff-team:
The core of this PR — reintroducing `functions.<tool>.<index>` deterministic IDs backed by a monotonic counter in `AgentState.toolCallState`, seeded from history (including the legacy colon-delimited shape) and pending calls — is well done. `tool-call-id.ts` is clean, `createToolCallIdGenerator`/`ensureToolCallState` correctly thread the counter through spawned agents, and the new test file (`tool-call-id.test.ts`) plus updates to `run-programmatic-step.test.ts` and `tool-validation-error.test.ts` cover the important cases (history seeding, hidden/legacy IDs, shared counter across spawned agents).

What's concerning is that `packages/agent-runtime/src/tools/tool-executor.ts` also contains a second, unrelated change to agent-spawn validation that has nothing to do with tool call IDs:

- `agentIdToLoad` is now normalized via `normalizeAgentIdForLookup` for base agents, and the returned `agent` object is rewritten with `agent_type: agentIdToLoad`.
- The `effectiveInput = { ...effectiveInput, agents: validAgents }` line was moved outside the `if (errors.length > 0)` block, so it now runs unconditionally instead of only when some agents failed to spawn.

Neither of these is mentioned in the PR description, and there's no test covering the new unconditional overwrite of `effectiveInput.agents`. That's a real behavior change (previously a successful spawn left `effectiveInput` untouched; now it always gets rebuilt from `validAgents`), and mixing it into a PR titled 'Restore state-backed tool call IDs' makes it easy for a reviewer to miss and hard to isolate for porting.

Please split the spawn-agent normalization change into its own PR with its own tests and description so each change can be evaluated (and reverted, if needed) independently. The tool-call-id work itself looks portable once separated.

PR #622, opened 2026-05-08
Use DeepSeek agent for BuffBench

jahooma:
## Summary
- Switch BuffBench main runner from `base2-free-evals` to `base2-free-deepseek`.

## Validation
- Reviewed `git diff origin/main...HEAD`.


codebuff-team:
The stated change — swapping `base2-free-evals` for `base2-free-deepseek` in `evals/buffbench/main.ts` — is a one-line, sensible change and clearly in scope.

However the diff is much larger than the description suggests. `agents/base2/base-deep.ts` and `agents/base2/base2.ts` also drop `file-picker`, `file-picker-max`, `directory-lister`, and `glob-matcher` from `spawnableAgents`, rewrite the system/step prompts to point at `glob`/`list_directory`/`read_subtree`/`read_files` instead, and get a bunch of unrelated prettier-style reformatting (multi-line function signatures, template literal indentation). None of that is mentioned in the PR body, and there's no rationale given for retiring those sub-agents or reason to believe it's tied to the DeepSeek swap.

As it stands this reads like two unrelated changes squashed together: a trivial eval config change, and a meaningful behavioral change to how base2/base-deep orchestrate sub-agents. The latter deserves its own PR with a description of why those agents are being removed and what testing (if any) shows the direct-tool approach performs as well. Please split these so each can be reviewed and ported on its own merits.

PR #667, opened 2026-05-13
Remove freebuff waiting room flag

jahooma:
Removes the Freebuff waiting-room feature flag now that the session gate is permanently enabled. The server always starts admission, no longer emits disabled gate responses, and no longer fetches user email for the removed bypass path. CLI compatibility for older servers is preserved with the client-side disabled fallback, while docs and tests now describe the always-on session gate. Validated with package typechecks, focused Freebuff/session tests, chat completions tests, admission tests, and git diff --check.


codebuff-team:
This PR removes the `FREEBUFF_WAITING_ROOM_ENABLED` flag now that the session gate is always on, and the change itself looks clean and well-tested — the CLI-side compatibility fallback for older servers, the doc updates in `docs/freebuff-waiting-room.md`, and the removal of the `team@codebuff.com` bypass and unused `userEmail` plumbing all read as a coherent, deliberate cleanup with corresponding test removals rather than blind churn.

However, per this mirror's scope rules, changes to `web/` and `packages/internal/` are not accepted here regardless of quality. The bulk of the substantive logic changes are in `web/src/server/free-session/*`, `web/src/app/api/v1/freebuff/session/*`, `web/src/app/api/v1/chat/completions/*`, and `packages/internal/src/env-schema.ts` — all out-of-scope paths for this repo. Only the CLI-side comment tweaks and doc changes would be in-scope, and those are cosmetic without the server-side change they describe.

If you want this ported, it needs to go through the private repo directly since the core of the diff can't land here. Nothing wrong with the work, just the wrong venue for the parts that matter.

PR #695, opened 2026-05-16
fix: respect http_proxy env vars in CLI API client

shabeer:
Bun's native fetch does not automatically honour HTTPS_PROXY / http_proxy environment variables unlike Node's http/https modules. Read proxy settings from standard env vars and pass them via Bun's fetch  option so corporate proxy users can route CLI traffic.

### DIFF
--- a/cli/src/utils/codebuff-api.ts
+++ b/cli/src/utils/codebuff-api.ts
@@ -103,6 +103,13 @@ export interface CodebuffApiClientConfig {
   defaultTimeoutMs?: number
   /** Default retry configuration */
   retry?: RetryConfig
+  /**
+   * Proxy URL to use for all requests.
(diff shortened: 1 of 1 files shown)


shabeer:
This possibly fixes the issue in https://github.com/CodebuffAI/codebuff/issues/694

shabeer:
@jahooma @brandonkachen @charleslien - Any updates on the code review?

jahooma:
This might work for some api requests, but a lot are made through the sdk/agent-runtime packages. 

I'll merge this for now, but the whole proxy thing needs to be through through. We might just want to delete it if is tricky to make it actually work.

shabeer:
Thanks @jahooma for merging the PR

PR #745, opened 2026-05-24
Route deepseek-v4-flash to Fireworks when DeepSeek API is unhealthy

jahooma:
## Summary
- Adds Fireworks as a transparent fallback for `deepseek-v4-flash` (`accounts/fireworks/models/deepseek-v4-flash`).
- New passive circuit breaker (`deepseek-health.ts`): **3 failures / 60s window opens the circuit for 5 min**, then the next request probes DeepSeek and resets on success. No background polling — every user request is the probe, so all pods converge naturally.
- Tighter **60s `headersTimeout`** for the Flash undici agent so dead-API requests fail fast instead of hanging on the existing 30-min default (kept for reasoning models on v4-pro).
- `_post.ts` routes to Fireworks when the circuit is open, plus inline pre-stream failover so the **first** user to hit an outage also gets a Fireworks response instead of an error.
- Pricing entry in `FIREWORKS_PRICING_MAP`: 0.14 / 0.03 / 0.28 per M tokens (input / cached / output).

## How it works
1. `createDeepSeekRequestTracked` wraps the DeepSeek fetch. Network errors, timeouts, 5xx/408/429 → `recordDeepSeekFailure()`. 2xx → `recordDeepSeekSuccess()` (clears state).
2. When `recentFailures.length >= 3` within the 60s window, `openUntil = now + 5min`.
3. Routing in `_post.ts` calls `shouldBypassDeepSeek(model)` and, if true, sets `useDeepSeek = false` so the existing Fireworks branch picks up the same model id (now in `FIREWORKS_MODEL_MAP`).
4. After cooldown expires, the next request retries DeepSeek directly. Success resets; another failure re-opens.

## Test plan
- [x] Unit tests for circuit breaker and outage classifier (11 new tests in `deepseek-health.test.ts`, all passing).
- [x] Existing `fireworks-deployment` and `fireworks-health` test suites still pass (75 llm-api tests total).
- [x] `bun run typecheck` clean for changed files (pre-existing SDK errors unrelated).
- [ ] Manual verification once deployed: confirm `deepseek/deepseek-v4-flash` calls succeed via DeepSeek when healthy and via Fireworks once the breaker opens (induce by temporarily pointing `DEEPSEEK_BASE_URL` at a sink, or watch real outage logs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


codebuff-team:
This is a well-structured change — a passive circuit breaker with sensible failure/window/cooldown parameters, tests for the classifier and breaker state transitions, and a reasonable fallback wiring in `_post.ts` — but every file touched (`web/src/app/api/v1/chat/completions/_post.ts`, `web/src/llm-api/deepseek-health.ts`, `web/src/llm-api/deepseek.ts`, `web/src/llm-api/fireworks.ts`, and the new test file) lives under `web/`, which is explicitly out of scope for this mirror per repo policy.

Since this is a member submission, the substance looks portable: the circuit-breaker design (in-process state, no polling, request-as-probe) and the tightened `headersTimeout` for the Flash agent are both defensible engineering choices. If this needs to land, it should be applied directly to the private source tree rather than merged through the public mirror.

No action needed here other than closing as out-of-scope for this repo; happy to review a version of this diff scoped to non-`web/` shared packages if there's a portable piece (e.g., the classifier logic in `deepseek-health.ts` if it were relocated to a shared package).

PR #747, opened 2026-05-25
Fix code search context line parsing

hiSandog:
## Summary
- Parse code_search context lines using known match file paths so hyphenated paths like src/component-2-test.ts are not truncated
- Preserve context content that also contains hyphen-number-hyphen sequences
- Add regression coverage for both cases

## Validation
- git diff --check
- Attempted: ./.bin/bun test common/src/util/__tests__/format-code-search.test.ts (blocked locally because the repository wrapper requires .env.local/Infisical and no real bun executable is installed on PATH)


hiSandog:
Closing this stale PR because it is older than 15 days. I am leaving the branch intact.

PR #762, opened 2026-05-27
Fix safeReplace to replace all matches

hiSandog:
## Summary
- update safeReplace to replace every matching occurrence instead of only the first
- add coverage for replacing multiple matches and keeping dollar signs literal in replacement text

## Validation
- env PATH=/Users/sandog/.real/.bin:/usr/local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin /Users/sandog/.real/.bin/bun test common/src/util/__tests__/string.test.ts

### DIFF
--- a/common/src/util/__tests__/string.test.ts
+++ b/common/src/util/__tests__/string.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'bun:test'
 
-import { pluralize } from '../string'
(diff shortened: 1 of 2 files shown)


hiSandog:
Closing this stale PR because it is older than 15 days. I am leaving the branch intact.

PR #805, opened 2026-06-18
fix: remove duplicate zod entry from root package.json

GautamKumarOffical:
zod was listed in both dependencies and overrides sections. Removed from dependencies since it's a workspace dependency that should only be in overrides to ensure consistent versions across all packages.

Simple cleanup to avoid confusion and potential version conflicts.


codebuff-team:
Good catch — having `zod` listed in both `dependencies` and `overrides` is redundant and can cause confusion about which version is authoritative. Since `overrides` already pins `zod` for the whole workspace, dropping it from `dependencies` is the right cleanup, assuming no code directly under the root package imports `zod` (worth double-checking, since removing it from `dependencies` means it's no longer a direct dependency, just a transitively-hoisted one via workspace packages).

Small, focused, low-risk change. Worth a maintainer verifying that root-level scripts/tooling don't import `zod` directly before porting, but otherwise straightforward.

victorxheng:

PR #814, opened 2026-06-26
fix: add linux-x64-baseline target to build script

### LINKED ISSUE #797: linux-x64-baseline target still contains AVX instructions, causing SIGILL on older CPUs (SSE4.1 only)
### Description                                                                                                                         
  The target `linux-x64-baseline` (to support older CPUs) is still failing with a `SIGILL` (illegal instruction) 
  on CPUs that do not support AVX or AVX2 instructions at all (such as older CPUs supporting only up to SSE4.1/SSE4.2).                   
                                                                                                                                          
  ### Technical Details                                                                                                                   
  Analyzing the downloaded `linux-x64-baseline` binary shows that it was compiled with AVX vector instructions (utilizing `%ymm` 
  registers), which causes the crash on processors without AVX support.                                                                 
                                                                                                                                        
  **Example instructions found in the `linux-x64-baseline` binary:**
  ```assembly
  405739c:    c5 dd eb eb            vpor   %ymm3,%ymm4,%ymm5
  40573a0:    c5 ed eb f1            vpor   %ymm1,%ymm2,%ymm6

  CPU Flags of the affected system:
  flags: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ht tm pbe syscall nx
  lm constant_tsc arch_perfmon pebs bts rep_good nopl cpuid aperfmperf pni dtes64 monitor ds_cpl vmx est tm2 ssse3 cx16 xtpr pdcm sse4_1
  xsave lahf_lm pti tpr_shadow flexpriority vpid dtherm vnmi
  (No avx or avx2 flags present).

  Expected Behavior

  The linux-x64-baseline binary should be compiled targeting a generic x86-64 architecture (or at most sse4.2 / sse4a level
  compatibility), entirely avoiding AVX instructions and %ymm register utilization, so that it can run on older systems as intended.
  ```

GautamKumarOffical:
The linux-x64-baseline target was missing from the build script's target mappings, preventing the creation of a baseline binary for older CPUs that don't support AVX instructions.

This adds the linux-x64-baseline mapping so the build can produce a binary compatible with CPUs that only support SSE4.1/SSE4.2.

The runtime already has AVX2 detection and baseline fallback logic (index.js lines 225-248), but the build script couldn't produce the baseline binary because the target wasn't mapped.

Fixes #797


codebuff-team:
Good, focused fix. The build script's `mappings` object was missing the `linux-x64-baseline` entry, which lines up with `bun-linux-x64-baseline` being a real Bun compile target and with the runtime's existing AVX2/baseline detection logic in index.js (lines 225-248) that you cite. This is a plausible root cause for #797 - without this mapping, `getTargetInfo()` would throw or fall through for that target, so the baseline binary could never be produced.

A few things worth checking before this is ported: does `platform`/`arch` need any other distinguishing field (e.g. a `baseline: true` flag) so downstream code that consumes `TargetInfo` can tell this apart from regular `linux-x64` for packaging/upload purposes? If `TargetInfo` only has `bunTarget`/`platform`/`arch`, and other code branches on `platform`+`arch` to decide filenames or artifact names, both `linux-x64` and `linux-x64-baseline` would collide. Worth a quick look at how `TargetInfo` is consumed elsewhere in the file.

No tests are added, but this is a small data/config change, so that's likely fine for a script like this. Overall this looks correct and small enough to review confidently - a solid one-line-cause, one-mapping-fix PR.

mVRx3i:
Hi @codebuff-team and @GautamKumarOffical!

In the client index.js, linux-x64-baseline is already mapped to freebuff-linux-x64-baseline.tar.gz.

Once this PR is merged and the build server compiles the baseline binary without AVX, it will fix the SIGILL crash on older CPUs.

Thanks for reviewing this fix!

victorxheng:

PR #829, opened 2026-06-30
docs: add Windows baseline troubleshooting

vamshikrishnaramasamy:
Documents the Windows illegal instruction / older CPU workaround for Codebuff and Freebuff.

This adds steps for forcing the win32-x64-baseline binary, clearing existing metadata, and verifying the installed target.\n\nReferences #822.


codebuff-team:
Nice addition. The symptom (SIGILL / exit code 0xC000001D) matches a real class of bugs with prebuilt native binaries that assume AVX2, and the format matches the rest of WINDOWS.md (Symptom/Cause/Solutions/Reference), which makes it easy to fold into the doc as-is.

A few things worth double-checking before this gets ported over:

1. Confirm `FREEBUFF_BINARY_TARGET` / `CODEBUFF_BINARY_TARGET` are the actual env var names used by the installer/updater logic — if the real names differ, this will send users down the wrong path.
2. Confirm the cache path is really `%USERPROFILE%\.config\manicode\...` for both freebuff and codebuff on Windows (rather than something under `%APPDATA%` or `%LOCALAPPDATA%`, which is more idiomatic for Windows config storage). If the actual path differs per-binary, say so explicitly.
3. Minor: PowerShell users will need `$env:VAR=...` and `Remove-Item`, not `set`/`del` — you show a PowerShell prompt in the Symptom block but cmd.exe syntax in the Solutions block. Worth adding both or noting cmd.exe explicitly.

Assuming the env var names and paths are verified against the actual binary-resolution code, this is a low-risk, useful docs change.

victorxheng:

PR #840, opened 2026-07-03
fix: use canonical www.codebuff.com URL for fallbacks

OmarAlaaeldein:
This PR fixes a silent authorization header-stripping issue during local development/testing.

- Change hardcoded fallback URLs from `https://codebuff.com` to the canonical `https://www.codebuff.com`.
- This prevents the 301 redirect from `codebuff.com` to `www.codebuff.com` which causes Node's `fetch` client to strip the `Authorization` header for security reasons, resulting in a silent `401 Unauthorized` error.


codebuff-team:
Good catch and a minimal, focused diff. Using `https://codebuff.com` as a fallback when it 301-redirects to `www.codebuff.com` is a real footgun for `fetch`, which drops `Authorization` headers on cross-origin redirects per the Fetch spec's redirect handling for sensitive headers. Switching the fallback to the canonical `www.codebuff.com` in `cli/src/hooks/use-freebuff-session.ts` and `cli/src/utils/error-handling.ts` avoids the redirect entirely, and the doc-comment update in `sdk/src/validate-agents.ts` keeps the example consistent.

A few things worth checking before porting:
- Confirm `https://www.codebuff.com` is actually the domain that serves `/api/v1/...` — if the API is hosted only on the bare domain and www is just a marketing redirect, this could break things instead of fixing them. Worth a quick curl/test against both endpoints.
- No test coverage is added, though given the size and nature (a literal fallback string swap) that's understandable and probably not blocking for a maintainer to port by hand.
- It would help to also check if there are other places in the codebase using the same `codebuff.com` fallback pattern (e.g. web/ or shared config) that weren't touched here, even though those are out of scope for this mirror — a maintainer porting this should search for consistency.

Overall this is small, well-reasoned, and easy to verify/port by hand.

OmarAlaaeldein:
Rebased onto current main. The session endpoint moved to cli/src/utils/freebuff-session-api.ts, so the canonical fallback is applied there along with the error-handling fallback and SDK examples. Verified both endpoints: the bare domain redirects to https://www.codebuff.com/api/v1/freebuff/session, and the canonical endpoint serves that route (both correctly return 401 without credentials). I also searched the remaining NEXT_PUBLIC_CODEBUFF_APP_URL fallbacks; the only other one is an unauthenticated release-download URL, so it is outside this Authorization-header redirect issue.

victorxheng:

PR #846, opened 2026-07-05
feat(ops): scheduled peak-hours scaling for the Render web service

jahooma:
## Problem

The `web` service (the API/completions backend — **not** freebuff/web) runs a **flat 6 instances 24/7 with no autoscaling**, while load swings ~2x on a predictable daily curve. From the 2026-07 observability work (`event_loop_lag`, `chat_completion_concurrency`):

| | off-peak (02–07Z) | peak (12–17Z) |
| --- | --- | --- |
| event-loop `utilization` avg / max | ~0.40 / 0.62 | **0.73–0.76 / 0.93** |
| max concurrent chat requests | ~43 | ~86 |

The same 6 pods absorb the whole swing (`dcount(host)` per hour is exactly 6). There's **no `render.yaml`** in the repo — the service is dashboard/API-managed. Service id `srv-crm8estds78s73e7evd0` (derived from prod pod hostnames).

## Approach — scheduled instance floor

Load is clock-predictable and the bottleneck (event-loop saturation → ingress queue) **leads** container CPU, so Render's CPU-target autoscaling reacts late. A time-based floor is more precise. This drives `numInstances` on a per-UTC-hour curve via Render's manual scale endpoint.

## The curve (re-derived over 14 days)

**Data caveat:** the utilization metric only went live ~07-03 23Z, so there's exactly **one full util-day (Sat 07-04) — which ran ~65% above normal load.** Sizing straight off it over-fits that spike (peaks at 14). Instead we calibrate `util ≈ 0.115 + 5.2e-5·load` from 07-04 and project onto the *normal* 14-day hourly load shape (`scripts/ops/derive-scale-curve.ts`). A typical day peaks at **7–8**:

```
00–04: 6   05–07: 7   08: 8   09: 7   10: 8   11–14: 7   15–17: 8   18–22: 7   23: 6
```

- **Weekday vs weekend: negligible** — total load within ~2%, per-hour counts differ by ≤1, so a single curve is used.
- **Trade-off:** this sizes for *normal* load. On a normal day the flat-6 fleet already only hits ~0.50 util; the worst queueing we saw was the 07-04 spike. Spike days will still see elevated queue at this floor — for spike protection, raise the peak or use native autoscaling (documented as Strategy B).

## Setup required to activate

1. Add repo **secret `RENDER_API_KEY`** (Render API key with access to the web service).
2. Optionally set repo **variable `RENDER_WEB_SERVICE_ID`** (defaults to the prod web service).
3. **Native autoscaling must be OFF** on the service — `POST /v1/services/{id}/scale` is silently ignored when autoscaling is enabled. The two strategies are mutually exclusive (see `docs/web-scaling.md`).

Until `RENDER_API_KEY` is set, the workflow is a **safe no-op** (skips, like the referral/bot sweeps).

## Files

- `scripts/ops/scale-web.ts` — the scaler (has `--dry-run`, `--instances`, `--hour`).
- `scripts/ops/derive-scale-curve.ts` — reproducible curve derivation from Axiom.
- `.github/workflows/web-peak-scale.yml` — hourly cron (idempotent, self-correcting) + `workflow_dispatch` override.
- `scripts/logs/web-health.ts` — reusable web-saturation probe (`bun scripts/logs/web-health.ts 24h`).
- `docs/web-scaling.md` — full write-up, both strategies, validation steps.

## Validation

`scale-web.ts` verified end-to-end in `--dry-run` across the curve; `derive-scale-curve.ts` run against 14d of prod Axiom data.

## Follow-up (separate PR)

A scheduled Axiom **saturation alert** (queue p95 / utilization thresholds) as a backstop — kept out of this PR to keep it focused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jahooma:
Opened against the public mirror by mistake (wrong remote). Re-opening against CodebuffAI/freebuff-private.

PR #853, opened 2026-07-07
fix(freebuff): derive commit summaries from git changes

victorxheng:
## User request

A Freebuff cloud user reported that commits shown in the Git tab after pushing use the original prompt text as the commit description instead of a meaningful summary of the actual project changes. The attached screenshot showed prompt-derived commit titles such as `Freebuff: Add this below the AI instructions...` and `Before: When committing changes...`.

## Implementation

- Added `summarizeGitChangesForCommit`, a deterministic shared helper that builds concise commit titles from git status/diff paths instead of user prompt text.
- Covered the reported case with a unit test: modified `AGENTS.md` plus new `CHANGELOG.md` now summarizes as `Update AI agent instructions and add changelog`.
- Strengthened the commit tool guide so agent-created commits explicitly base title/body on actual git status/diff and never copy/truncate the original prompt.
- Updated the example git-committer agent instructions to follow the same rule.

## Validation

- `bun test common/src/util/__tests__/git-commit-summary.test.ts`
- `cd common && bun run typecheck`
- `git diff --check`

## Confidence

Ready for review. The exact cloud Git-tab route is not present in this public snapshot, so this PR places the fix in shared code and the commit-generation contract that cloud/private code can call or inherit. The covered behavior directly matches the user report and prevents prompt-only summaries in the shared agent commit path.


codebuff-team:
The prompt-string edits in `run-terminal-command.ts` and `02-intermediate-git-committer.ts` are sensible, low-risk clarifications telling the agent to derive commit messages from git status/diff rather than the user's prompt. Those are fine as-is and easy to port.

The core of the PR, `summarizeGitChangesForCommit` in `common/src/util/git-commit-summary.ts`, is the problem. It's 175 lines of new logic plus 40 lines of tests, but nothing in this diff calls it — it's exercised only by its own test file. The PR description confirms this directly: "The exact cloud Git-tab route is not present in this public snapshot, so this PR places the fix in shared code... that cloud/private code can call or inherit." That means this doesn't actually fix the reported bug (prompt text showing up as commit titles in the cloud Git tab); it's a speculative function sitting in `common/src/util/` waiting for something to call it, which nothing here does. The reported issue is almost certainly in the push/commit-summary path used by cloud, not in the agent's `run_terminal_command` git-commit flow this PR touches — those are two different mechanisms (one is an LLM writing a commit message via the tool prompt, the other is presumably a non-LLM summary generated at push time).

As it stands, a maintainer porting this would get dead code with no call site, plus two small prompt tweaks that may or may not be relevant to the actual bug. I'd split this: drop the unused helper/tests, or find the actual call site in the private repo and wire it in with a real integration point and test that exercises the actual bug path, not a synthetic one.

victorxheng:

PR #872, opened 2026-07-15
feat(base2): add JavaScript Control-Flow Verification Rule to system prompt...

chicagoist:
prompt...

# Pull Request: Add JavaScript Control-Flow Verification Rule to Buffy Agent

---

**Repository:** [CodebuffAI/codebuff](https://github.com/CodebuffAI/codebuff)
**Target file:** `agents/base2/base2.ts`
**Target section:** `systemPrompt` → `# General guidelines`
**Branch:** `main`

---

## Title

```
feat(base2): add JavaScript Control-Flow Verification Rule to system prompt
```

---

## Problem

When debugging JS in HTML files, the agent falls into a **"syntax-blind loop"**:
it sees `node --check` fail, fixes brace imbalances mechanically, and declares
success when the parser passes — without verifying the code actually **executes**.

<details>
<summary>Real-world case study (30-round debugging cycle)</summary>

1. Agent ran `node --check` → SyntaxError (missing braces)
2. 10+ rounds of string-replacement fixes
3. `node --check` passed → agent declared "fixed"
4. **All interactive features were still broken**

Root cause: quiz logic was wrapped in `if (boxes.length > 0)` where `boxes =
querySelectorAll('.answerbox[contenteditable]')`. Those elements were removed
during a UI migration to checkboxes. `boxes.length` was always 0. Code was
syntactically valid but **logically dead**. The agent had grep, file reading,
and DOM analysis — but no mandatory checklist to use them.

</details>

**This is a recurring class of bugs.** The fix is not a new tool — it's a
mandatory methodology encoded in the system prompt.

---

## Proposed Change

One new bullet in `# General guidelines`, placed **after the closing backtick of
the Composio conditional block** (`${ENABLE_COMPOSIO_TOOLS ? ... : ''}`) and
**before the literal line `# Spawning agents guidelines`**.

### Unified diff

```diff
--- a/agents/base2/base2.ts
+++ b/agents/base2/base2.ts
@@ systemPrompt template literal, General guidelines section @@
 - **Discover and install skills:** ...
 ${ENABLE_COMPOSIO_TOOLS
   ? `...`
   : ''
 }
+- **JavaScript Debugging:** After fixing JS syntax errors, do NOT stop at
+  \`node --check\` passing — the code may be syntactically valid but logically
+  dead. Verify: (1) every \`querySelector(All)\` target exists in the HTML body,
+  (2) execution reaches every \`addEventListener\` (trace from IIFE entry through
+  all wrapping \`if\` conditions), (3) no \`if (elements.length > 0)\` guard wraps
+  live logic around removed DOM elements. Only after all three checks pass should
+  the fix be considered complete.
 # Spawning agents guidelines
```

### Rationale for single bullet

Existing guidelines use compact `- **Topic:** Actionable instruction` format.
This follows the same style. The three sub-checks are inline for brevity while
remaining specific enough to execute.

---

## The Three Rules (reference)

| # | Rule | Check |
|---|------|-------|
| 1 | **DOM cross-reference** | `grep '<selector>' *.html` → at least 1 match must exist |
| 2 | **Execution-path trace** | From IIFE entry to each `addEventListener` — every wrapping `if` must be reachable |
| 3 | **`boxes.length > 0` pattern** | Any guard wrapping >10 lines of logic is a red flag after UI migration |

---

## Contribution Checklist

- [x] Change scoped to `agents/` (per CONTRIBUTING.md)
- [x] No backend, database, billing, deployment, or secret-management code
- [x] Follows existing code style (template literal, bullet format)
- [x] Single bullet, minimal change — no structural refactoring
- [x] Applies to all modes (free/default/max/fast) — rule is in shared `systemPrompt`

---

## Alternatives Considered

| Approach | Verdict |
|----------|---------|
| New `instructionsPrompt` section | ❌ Too verbose, breaks existing flow |
| Separate skill file | ❌ Must be loaded manually each session |
| **Single bullet in Guidelines** | ✅ Always active, minimal, consistent |

---

## License

By submitting this pull request, I confirm that my contribution is made under
the terms of the project's license.


codebuff-team:
Thanks for the detailed writeup and the root-cause case study — that's a real class of bug (syntactically valid but logically dead code due to stale selectors) and worth thinking about.

That said, I'd hold off on porting this as-is:

1. **Scope mismatch**: `base2.ts`'s `systemPrompt` is the general-purpose guideline set used across all agent tasks, not just JS/HTML debugging. Adding a JS-specific, DOM-specific checklist here bakes a narrow heuristic into every prompt, for every language and task. This bloats the prompt for the ~95% of sessions that never touch a `querySelector`.

2. **Overfit to one anecdote**: the rule is derived from a single 30-round debugging session. It's plausible this generalizes, but there's no evidence here that this wording measurably improves agent behavior versus just adding tokens. Prompt changes like this really want before/after eval results (e.g., against a small suite of "looks fixed but isn't" repro cases), not just a plausible-sounding fix.

3. **Placement/mechanics**: the diff correctly targets the general guidelines section and matches the existing bullet style, so mechanically it's fine — no issue there.

4. **Better fix location**: if this genuinely recurs, a more targeted approach would be a specialized JS-debugging skill/agent invoked when doing HTML/JS work, rather than a permanent guideline in the base prompt.

If you have eval numbers or several repro cases showing this rule reduces false 'fixed' declarations, that would make a much stronger case for landing something like this — possibly as a narrower, generalized rule (e.g., 'don't declare a fix complete based on static checks alone; verify the code path executes') rather than DOM/JS-specific mechanics.

codebuff-team:
Marking this stale - there has been no activity here for 39 days. It will close in 7 days unless someone comments.

victorxheng:

PR #876, opened 2026-07-17
feat(cli): add read-only /mcp status command

fortissolucoescontato-bit:
## Summary

Adds a read-only `/mcp` command for inspecting configured MCP servers and the current runtime connection state.

The implementation reuses the existing MCP configuration loader and runtime state. It does not introduce a second MCP client, modify MCP configuration files, or initiate connections when rendering the status.

## Commands

- `/mcp`
- `/mcp list`

Unknown subcommands display inline usage information.

## Output

The command displays:

- configured server name;
- connection status;
- transport;
- discovered tool count;
- sanitized connection errors;
- sanitized configuration path.

When no servers are configured, it provides a short example using the existing `.agents/mcp.json` format.

## Security

- Read-only command;
- does not modify `.agents/mcp.json`;
- does not initiate MCP connections;
- redacts tokens, API keys, credentials and authorization headers;
- sanitizes credentials embedded in URLs;
- truncates oversized errors;
- replaces the user home directory with `~`;
- covers Linux, macOS and Windows path formats.

## Implementation

- Adds a public read-only MCP runtime status API in `common`;
- adds the `/mcp` command handler and formatter in `cli`;
- registers `/mcp` in the command registry and slash-command autocomplete;
- documents `/mcp` in the public README.

## Verification

- MCP command tests: **51** passing;
- related CLI tests: **217** passing (13 files);
- parser integration tests: PASS;
- lifecycle tests: PASS;
- SDK build: PASS;
- Freebuff build: PASS;
- local CI (`bun run ci`): PASS;
- typecheck: no new errors vs upstream/main;
- `git diff --check`: PASS.

## Scope

This PR intentionally does not implement:

- `/mcp add`;
- `/mcp remove`;
- `/mcp reload`;
- `/mcp inspect`.

Those can be considered separately after the read-only status command is reviewed.


codebuff-team:
Nice feature — a read-only `/mcp` status command that reuses existing config loading and connection state instead of spinning up a second MCP client is exactly the right shape for this. The `sanitizeErrorForDisplay`/`truncateError` additions in `common/src/mcp/client.ts` are a genuinely useful addition beyond just this command, and the test coverage in `mcp-command.test.ts` is thorough (redaction patterns, truncation boundaries, cross-platform path sanitization).

A few things worth fixing before porting:

1. `cli/src/components/help-banner.tsx` — the diff adds a stray leading space to the existing "Use @ to reference..." line (`-              Use @` → `+               Use @`, one extra space) and the new `/mcp` line has the same extra-space typo. Looks like an accidental find/replace artifact; please fix the indentation so it matches the surrounding `<text>` blocks exactly.

2. `connectionErrors` and `resolvedToolCounts` in `common/src/mcp/client.ts` are module-level maps keyed by config hash that are never cleared on success/reconnect (aside from the `connected` gate hiding stale errors). Bounded by number of distinct configs so not urgent, but worth a comment or TODO acknowledging the lifecycle.

3. The `SENSITIVE_PATTERNS` list is broad (e.g. `token=`, `secret=`) and could over-redact legitimate non-secret strings in server stderr — acceptable for a security-conservative default, but worth flagging in the PR description that false positives are expected.

None of these block review — the core design (read-only, no side effects, sanitized output) is sound and matches what a maintainer would want to see before adding write commands (`/mcp add`, etc.) later.

victorxheng:

PR #886, opened 2026-07-19
fix: prevent self-kill and ensure graceful shutdown with terminal flush

ilyaxuwu:
This should works now! https://streamable.com/r42l5x If it doesn't work I would give up


jahooma:
I'll have my agents take a look

codebuff-team:
Thanks for digging into the exit-cleanup path — the underlying idea (make sure buffered terminal escape sequences reach the tty before the process dies) is legitimate and worth pursuing. A few concerns before this is portable, though:

1. **`command-registry.ts`**: replacing `process.kill(process.pid, 'SIGINT')` with `process.exit(0)` changes behavior, not just timing. The SIGINT handler (via `use-exit-handler`) presumably drives the *async* cleanup path in `freebuff-exit.ts` (`endFreebuffSessionBestEffort`, `flushAnalytics`, `stopActiveRun`). `process.exit(0)` only fires the synchronous `'exit'` listeners, so the `exit`/`quit`/`q` command will now skip session-end and analytics flushing entirely. That's a regression, not a fix, unless those calls are also wired into the `'exit'` handler.

2. **`renderer-cleanup.ts`**: `process.stdout._handle?.setBlocking?.(true)` reaches into a private, untyped Node internal (`_handle` isn't on the `WriteStream` type), which will likely fail TypeScript compilation without an `any` cast and is fragile across Node/Bun versions. Also, reusing `handlersInstalled` (previously an installation guard) to also guard against double-`cleanup()` calls conflates two unrelated concerns and could silently disable handler re-installation elsewhere.

3. **`run-terminal-command.ts`**: the `child.pid === process.pid` guards in `killProcessGroup`/exit sweep protect against a scenario (pid collision between parent and its own child) that essentially can't happen under normal OS pid allocation. This reads as defensive code added without a concrete repro, which raises the question of whether the actual root cause of the terminal garbling was diagnosed.

The `freebuff-exit.ts` flush addition is the most clearly beneficial and low-risk piece here. I'd suggest splitting that out, dropping the self-kill guards unless you have a reproducible case, and reverting the `exit` command to keep the async cleanup path intact (or explicitly wiring cleanup into `process.on('exit')` if you want the synchronous route).

ilyaxuwu:
Oh yeah, by the way. I’d like your prompt to automatically check which PID is Freebuff is running on when it’s active, and when deciding to kill all instances of "node.exe" process, to check the PIDs in Task Manager and ensure that Freebuff’s own PID (for example, 15342) blacklist it before killing the pid. and then prevent Freebuff from shutting down.

codebuff-team:
Marking this stale - there has been no activity here for 35 days. It will close in 7 days unless someone comments.

victorxheng:

PR #887, opened 2026-07-20
fix: make CLI shutdown reliable

jahooma:
## Summary

- route all interactive and signal-driven exits through one idempotent cleanup coordinator
- destroy OpenTUI before the final synchronous controlling-terminal reset, while retaining the watchdog as a fallback
- drain active and buffered client-log batches within the bounded graceful-exit window
- remove unsupported stdout internals and speculative self-PID guards
- add focused coverage for cleanup ordering, log-drain concurrency, and exact terminal writes

## Root cause

Exit behavior was split across UI handlers, process signals, Freebuff cleanup, and renderer teardown. Some paths bypassed cleanup, buffered terminal writes could be lost or reordered after alternate-screen restoration, and an in-flight log request caused the exit flush to return without draining remaining records.

## Validation

- bun run build:sdk
- bun run --cwd cli typecheck
- bun run --cwd sdk typecheck
- 133 focused Bun tests
- real tmux runs for /exit, raw-mode double Ctrl+C, and SIGTERM, all exiting cleanly without TUI remnants


jahooma:
Closed in favor of the correctly targeted private-repository PR: https://github.com/CodebuffAI/freebuff-private/pull/828

Issue #912, opened 2026-07-29
MCP Tool  inputSchema.properties  stripped when registering tools from external MCP servers


billyandco:
## Environment

* **Runtime:** Freebuff (AI Agent Runtime)
* **Connected MCP Server:** `marionette_mcp` `v0.6.0` (by LeanCode)
* **Underlying Library:** `mcp_dart` `v2.3.0` / `v1.3.0`

---

## Symptom

When a third-party Model Context Protocol (MCP) server registers a tool with an `inputSchema` containing named `properties`, the Freebuff runtime registers the tool with `properties: {}`. As a result, the AI agent is unable to pass parameters to the tool.

### Example

The `connect` tool in `marionette_mcp` is defined as follows:

```dart
// marionette_mcp source (vm_service_context.dart)
server.registerTool(
  'connect',
  inputSchema: ToolInputSchema(
    properties: {
      'uri': JsonSchema.string(
        description: 'VM service URI (e.g., ws://127.0.0.1:8181/ws)...',
      ),
    },
    required: ['uri'],
  ),
  ...
);

```

This generates standard JSON Schema in the MCP `tools/list` response:

```json
{
  "name": "connect",
  "inputSchema": {
    "type": "object",
    "properties": {
      "uri": {
        "type": "string",
        "description": "VM service URI..."
      }
    },
    "required": ["uri"]
  }
}

```

However, Freebuff interprets the tool parameter schema as:

```json
{
  "parameters": {
    "type": "object",
    "properties": {},
    "additionalProperties": false
  }
}

```

---

## Impact

This affects **every tool requiring parameters**. Because `properties` is empty and `additionalProperties` is set to `false`, the AI agent cannot invoke any tool that accepts input arguments.

### Execution Error

Attempting to call `connect` results in:

```text
Invalid parameters for marionette__connect:
  path: ["uri"]  →  "expected string, received undefined"

```

> **Note:** The validation error path `["uri"]` indicates that a secondary validation layer correctly parses the schema (it knows `uri` is required), but the primary tool parameter definition exposed to the AI model retains `properties: {}`.

---

## Evidence

1. **Source Code Schema Definition:** `marionette_mcp` `v0.6.0` (`vm_service_context.dart`, lines 62–68) explicitly defines `properties: {'uri': JsonSchema.string(...)}` and `required: ['uri']`.
2. **Correct Serialization:** `mcp_dart`'s `JsonObject.toJson()` correctly outputs the properties:
```dart
return {
[...]

PR #921, opened 2026-07-31
fix(agent-runtime): preserve MCP tool schema properties

santhiprakash:
Problem
MCP tools registered through the runtime could lose nested JSON Schema details like `properties` before they reached the model-facing tool definition. That breaks parameterized tools such as `marionette__connect` from issue #912.

Triage / Root cause
`packages/agent-runtime/src/tools/prompts.ts` was round-tripping additional tool definitions through Zod and then re-emitting a sanitized schema. That conversion path was unnecessary for MCP tools and could discard schema shape that callers already supplied.

Fix
- Keep the original MCP JSON Schema object on the tool definition so nested fields survive registration unchanged.
- Continue using Zod conversion only for description generation.
- Add a regression test that asserts `properties.uri` survives `getToolSet()`.

Verification
- `/home/ubuntu/.bun/bin/bun test packages/agent-runtime/src/__tests__/prompts-schema-handling.test.ts`
- Result: pass (16 tests passed).

Notes / Risks
- I verified this against current `upstream/main` with the required preflight script before editing.
- The worktree had unrelated pre-existing edits in `sdk/src/tools/run-terminal-command.ts` and `sdk/src/__tests__/run-terminal-command.test.ts`; I left those untouched.
- No AI co-author trailer included.

Model Used
- gpt-5.4-mini

Checklist
- [x] Current upstream verified
- [x] Duplicate PR check passed
- [x] Targeted regression test passed
- [x] Commit authored as Santhi Prakash

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Instead of using one model for everything, Codebuff coordinates specialized agen
   <img src="./assets/codebuff-vs-claude-code.png" alt="Codebuff vs Claude Code" width="400">
 </div>
 
-Codebuff beats Claude Code at 61% vs 53% on [our evals](evals/README.md) across 175+ coding tasks over multiple open-source repos that simulate real-world tasks.
+Codebuff beats Claude Code at 61% vs 53% on [our evals](evals/buffbench/README.md) across 175+ coding tasks over multiple open-source repos that simulate real-world tasks.
 
 ## Freebuff: the free coding agent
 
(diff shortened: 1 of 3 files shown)


santhiprakash:
Verified the targeted regression test for this PR:\n\n- `/home/ubuntu/.bun/bin/bun test packages/agent-runtime/src/__tests__/prompts-schema-handling.test.ts`\n- Result: 16 pass, 0 fail\n\nNo additional code changes were needed in this pass.

chrismindpower369:
Nice root-cause analysis and fix. Preserving the original MCP JSON Schema should prevent parameterized tools from losing their input shape.

Could we add regression assertions for `required`, nested `properties`, `items`, and `enum` as well? The current `properties.uri` test covers the reported case, but these fields would make the protection more robust for custom MCP tools.

After resolving the merge conflicts, this looks good to me.

santhiprakash:
Closing: the source files modified by this PR (packages/agent-runtime/src/tools/prompts.ts) have been removed or reorganized on upstream/main. The patch no longer applies.

PR #927, opened 2026-08-01
fix: add GLM 5.2 to FREEBUFF_MODELS for referral-unlocked users

youlianvr-lgtm:
GLM 5.2 was intentionally excluded from FREEBUFF_MODELS because it was treated as a referral-only reward. However, freebuffModelOptions() adds GLM to the selector when glmUnlocked is true (user earned it via referrals), but isFreebuffModelId() checks FREEBUFF_MODELS which didn't include it — causing 'invalid model' error when the user tries to select the unlocked model.

Adding GLM_V52_MODEL to FREEBUFF_MODELS fixes the mismatch between the selector and validation logic. The referral system still controls whether the model appears in the UI via freebuffModelOptions().

### DIFF
--- a/common/src/constants/freebuff-models.ts
+++ b/common/src/constants/freebuff-models.ts
@@ -548,10 +548,10 @@ export const SUPPORTED_FREEBUFF_MODELS = [
   MIMO_V25_MODEL,
 ] as const satisfies readonly FreebuffModelOption[]
 
-// GLM 5.2 is intentionally NOT in FREEBUFF_MODELS: it isn't a freely-pickable
-// grid model, it's a referral reward surfaced by the separate referral banner.
-// It stays in SUPPORTED_FREEBUFF_MODELS so the session/chat layers accept it as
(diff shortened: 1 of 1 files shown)


youlianvr-lgtm:
So you understand, this is a problem for Freebuff Desktop.

common/src/constants/freebuff-models.ts
```diff
@@ -548,10 +548,10 @@ export const SUPPORTED_FREEBUFF_MODELS = [
   MIMO_V25_MODEL,
 ] as const satisfies readonly FreebuffModelOption[]
 
-// GLM 5.2 is intentionally NOT in FREEBUFF_MODELS: it isn't a freely-pickable
```
youlianvr, line 551: Honestly, looking back, I misspoke here. The main problem was the bug where the model was available in the selector, but trying to select it would throw an 'invalid model' error. I checked and realized that for this to work, it needs to be added to FREEBUFF_MODELS.

PR #966, opened 2026-08-10
fix(cli): reload MCP config after project selection

### LINKED ISSUE #957: [Bug] MCP servers from `.agents/mcp.json` are not loaded when the project is selected via the project picker (CLI started from an ancestor directory)
## Summary

When the Freebuff CLI is started **from a directory outside the project** and the project is then selected via the **project picker**, the MCP servers defined in the project's `.agents/mcp.json` are silently **not** exposed to the main agent. The `mcpServers` are only loaded if the CLI is started **directly inside the project directory** (e.g. `cd ~/projects/foo && freebuff`).

## Environment

- CLI: `freebuff` npm package `0.0.142` (native binary at `~/.config/manicode/freebuff`)
- OS: Linux
- Tested: 2026-08-07

## Steps to reproduce

1. Create a project with an `.agents/mcp.json` containing one or more MCP servers, e.g.:

   ```json
   {
     "mcpServers": {
       "context7": {
         "type": "http",
         "url": "https://mcp.context7.com/mcp",
         "headers": { "Authorization": "Bearer $CONTEXT7_API_KEY" }
       },
       "chrome-devtools": {
         "command": "npx",
         "args": ["-y", "chrome-devtools-mcp@latest", "--headless", "--isolated"]
       }
     }
   }
   ```

2. Start the CLI from a directory **above** the project (e.g. `cd ~ && freebuff`).
3. In the project picker, select the project.
4. Ask the main agent which tools it has available.

**Result:** the MCP tools (`context7__*`, `chrome-devtools__*`) are **not** present.

5. Now exit, and restart with `cd ~/projects/<project> && freebuff` (no project picker).
6. Ask the same question.

**Result:** the MCP tools **are** present.

## Expected behavior

MCP servers from the selected project's `.agents/mcp.json` should be available to the main agent regardless of which directory the CLI was launched from.

## Actual behavior

MCP servers are only available when the CLI is launched **inside** the project directory. Selecting the project afterwards via the picker never exposes them.

## Root cause analysis

The MCP config is loaded **once, at boot time**, using the then-current `process.cwd()`:

- `cli/src/index.tsx` (~line 260): `await initializeAgentRegistry()` runs during startup. This calls `loadMCPConfigSync({ verbose: false })` (`sdk/src/agents/load-mcp-config.ts`), which resolves `{cwd}/.agents/mcp.json` against `process.cwd()` and populates a **module-level** `mcpServersCache`.
- `cli/src/index.tsx` (~line 347): only **later**, when the user picks a project, does the CLI call `process.chdir(newProjectPath)` followed by `resetCodebuffClient()`.
- `resetCodebuffClient()` recreates the client but **never reloads** `mcp.json` — the `mcpServersCache` module variable was already populated (with an empty/absent config) at boot.

The merge itself is correct and does run afterwards:

- `cli/src/utils/local-agent-registry.ts` merges `mcpServersCache` into every agent definition whose `id.startsWith('base')` (the Freebuff root agent `base2-free-deepseek-flash` qualifies):

  ```ts
  for (const agent of definitions) {
    if (agent.id.startsWith('base')) {
      if (!agent.mcpServers) agent.mcpServers = {};
      agent.mcpServers = { ...agent.mcpServers, ...mcpServersCache };
    }
  }
  ```

  (Confirmed in the installed binary's minified JS: `if(Object.keys(VW).length>0){for(let D of A)if(D.id.startsWith("base")){if(!D.mcpServers)D.mcpServers={};D.mcpServers={...D.mcpServers,...VW}}}`.)

Because the cache was populated at boot from the **wrong** cwd, the merge operates on an empty cache → the main agent template ends up with `mcpServers` missing/null.

## Evidence

Session state (`run-state.json`) for the same chat, comparing both launch modes:

| Launch mode | Main agent template `mcpServers` |
|---|---|
| `freebuff` from `~`, project picked via picker | missing / `null` |
| `cd ~/projects/mathema && freebuff` | `{ "context7": …, "chrome-devtools": … }` |

The agent's registered `toolNames` (from the session log) match: `context7__resolve-library-id`, `context7__query-docs`, and 29 `chrome-devtools__*` tools appear only in the second mode.

## Suggested fixes

1. **Reload MCP config on project change** — after `process.chdir(newProjectPath)` in the project-picker path, re-run the MCP config load (or invalidate/re-populate `mcpServersCache` and re-run `loadAgentDefinitions()` / `initializeAgentRegistry()`).
2. **Resolve `mcp.json` from an explicit project root** — instead of `process.cwd()`, pass the resolved project path to `loadMCPConfigSync` so it is independent of launch directory and `chdir` timing.
3. **Load lazily** — resolve the MCP config from the project root at session/run start rather than at process boot, so the picker path is naturally covered.

## Workaround

Always launch the CLI from inside the project directory:

```bash
cd ~/projects/<project> && freebuff
```

## Affected files (approx., from `main`)

- `cli/src/index.tsx` — boot order: `initializeAgentRegistry()` before `process.chdir()` / `resetCodebuffClient()` in the project picker path
- `cli/src/utils/local-agent-registry.ts` — the merge of `mcpServersCache` into `base*` agent definitions
- `sdk/src/agents/load-mcp-config.ts` — `loadMCPConfigSync` resolves config from `process.cwd()`

luantaraschi:
## Summary

- Reload the local agent registry after the project picker changes the working directory, so the selected project's `.agents/mcp.json` is loaded before the client is recreated.
- Preserve the existing `--agent` override behavior.
- Add a regression test that verifies a project MCP server reaches the base agent definition after project selection.

Fixes #957

## Testing

- Project picker and local agent registry tests: 41 passed
- CLI typecheck
- SDK build
- Freebuff Windows binary build
- Freebuff version and help E2E tests: 5 passed


chrismindpower369:
This looks like the stronger fix for #957 because the regression test exercises the actual activateProject() flow and verifies that the selected project's MCP server reaches the base agent definition.

PR #973 appears to address the same root cause with overlapping changes. Before merging, could we confirm that this PR also clears every derived local-agent cache that depends on the previous cwd/project root? If yes, I would suggest merging #966 and closing #973 as a duplicate to avoid two competing fixes.

luantaraschi:
Thanks for catching this. The local agent registry has two derived caches that depend on the previous working directory: `cachedAgentsByMode` and `cachedAgentsDir`. The new `reloadLocalAgentRegistry()` helper clears both before `initializeAgentRegistry()` refreshes the agent definitions, file paths, and MCP configuration.

I also expanded the regression test to warm the caches in the launch project, switch projects through `activateProject()`, and verify that the selected project supplies the agent directory, local agent list, MCP server, and base agent definition. The old project's agent must no longer be present.

Validation on the new head (`2f246ab`):

- 41 related tests passed
- CLI typecheck passed
- Freebuff Windows binary build passed
- 5 `--version` and `--help` E2E tests passed
- Prettier and `git diff --check` passed

This incorporates the cache-reset part of #973 while keeping the project-picker coverage in this PR.

chrismindpower369:
Perfect, thanks for the thorough follow-up. Clearing both cwd-dependent caches and testing the full activateProject() transition addresses the concern. The added assertion that the old project's agent is absent is especially valuable.

This now seems like the right PR to merge for #957. I’d suggest closing #973 as a duplicate once a maintainer has reviewed this update.

codebuff-team:
Good catch and clean fix. The root cause is real: `handleProjectChange` in `cli/src/index.tsx` previously called `setProjectRoot`/`resetCodebuffClient` but never re-scanned `.agents/` for the newly selected directory, so `cachedAgentsDir`/`cachedAgentsByMode` in `local-agent-registry.ts` kept serving the launch directory's MCP config. Wrapping the chdir/setProjectRoot/registry-reload/reset sequence into a single `activateProject()` helper in `project-picker.ts` is a sensible consolidation and removes the duplicated logic from `index.tsx`.

The added test in `project-picker.test.ts` is a genuine regression test — it writes real `.agents/*.ts` and `mcp.json` files into temp launch/project directories, calls `activateProject`, and asserts the base agent definition picks up the new project's MCP server while dropping the launch project's local agent. That's exactly the kind of test this bug needed.

Two things worth double-checking before porting:
1. `reloadLocalAgentRegistry` only clears `cachedAgentsByMode` and `cachedAgentsDir` — please confirm there isn't a third cache (e.g. for MCP servers specifically, or directory-scan results) elsewhere in `local-agent-registry.ts` that also needs clearing, since `getLoadedMCPServers()` presumably reads from one of these.
2. The `hasAgentOverride` skip path (`reloadAgentRegistry: !hasAgentOverride`) isn't exercised by the new test — a second test case confirming the override path doesn't reload would strengthen this further, though the existing description of preserving that behavior is plausible from the diff.

No forbidden paths touched, no scope creep — solid, focused bug fix.

luantaraschi:
Both checked.

On the first one, there is no third cache. The module keeps five mutable ones: `userAgentsCache`, `userAgentFilePaths` and `mcpServersCache` at the top, and `cachedAgentsDir` and `cachedAgentsByMode` further down. `initializeAgentRegistry()` reassigns the first three on both the success and the failure path, including `mcpServersCache` from `loadMCPConfigSync()`, which reads the cwd, so `getLoadedMCPServers()` is covered as long as the chdir happens first. That is why `reloadLocalAgentRegistry()` only clears the other two: they are derived and nothing else resets them. `__resetLocalAgentRegistryForTests` lists all five, which is a handy way to confirm the count did not grow.

On the second one, you were right that the skip path had no coverage. Added a test for it: with `{ reloadAgentRegistry: false }` the chdir and project root still move, and the launch project's agent is still the one served. I checked it is worth something by making `activateProject` reload unconditionally, and the new test is the only one that fails.

victorxheng:

luantaraschi:
No problem, and thanks for explaining what happened rather than leaving it to look like a silent close.

Recreated as #1179, rebased onto `d4902003`. Same change, with the two follow-ups from this thread already folded in: the cache count and the skip-path test.

PR #978, opened 2026-08-12
fix(llm-providers): pass complete data URLs through image conversion

### LINKED ISSUE #977: [BUG] preview_screenshot crashes the turn with "Invalid image_url" on vision-capable models
On a vision-capable model (ChatGPT-5.6 Luna/Mimo), calling `preview_screenshot` crashes the turn
with an invalid image payload. The screenshot tool itself returns "ok", but the image attached
to the model request is not a valid base64 data URL:

    ⚠️ Provider returned error [OpenAI: { "error": { "message":
    "Invalid 'input[126].content[0].image_url'. Expected a base64-encoded data URL with an image
    MIME type (e.g. 'data:image/png;base64,...'), but got an invalid base64-encoded value.",
    "type": "invalid_request_error", "param": "input[126].content[0].image_url",
    "code": "invalid_value" } }]

## Key observation

- The crash happens ONLY on models that DO support vision
- Text-only models are NOT affected: the orchestrator does not attach the screenshot to them,
  so there is nothing to reject.
- Conclusion: the bug is in the screenshot pipeline (binary → base64 encoding / transport).

## Steps to reproduce

1. Start a conversation on a vision-capable model (e.g. ChatGPT-5.6 Luna/Mimo 2.5).
Then the agent should:
2. Register a preview (htmlPath or url+pid).
3. Call `preview_screenshot`.
--- 
4. The tool returns "ok", but the next model turn fails with the invalid image_url error above.
5. Repeats on every retry in this session.

## Environment

- OS: Windows 11 LTSC IoT Enterprise
- Freebuff Desktop 0.55
- Preview feature: `register_preview` + `preview_screenshot`
- Affected model: ChatGPT-5.6 Luna/Mimo-2.5. Text-only models unaffected.

---

I used Freebuff to help me write this issue.

youlianvr-lgtm:
## What

Screenshot tools can hand the OpenAI-compatible converter a **complete** `data:<mime>;base64,...` URL instead of raw base64. The converter in `packages/llm-providers` unconditionally wrapped every string value in a second `data:<mime>;base64,` prefix, producing payloads like:

```
data:image/png;base64,data:image/png;base64,iVBORw0KG...
```

The nested prefix injects `:` and `,` characters where providers expect base64, so requests fail with:

```
Invalid 'input[126].content[0].image_url'. Expected a base64-encoded data URL with an image MIME type, but got an invalid base64-encoded value.
```

Vision-capable models are affected because media is only attached to them — text-only models never receive an image part, so the bug hides.

## Fix

Detect a leading `data:` prefix in string data and pass the value through untouched. Raw base64 strings, `Uint8Array`, `ArrayBuffer`, `Buffer`, and `URL` inputs behave exactly as before (existing tests unchanged).

## Why here

The producing tool lives in the private client; the converter is the public chokepoint every OpenAI-compatible provider path goes through, so hardening it fixes the whole class of double-wrapped media regardless of which tool produced the data.

## Testing

- New regression test: an already-complete data URL is emitted unchanged (no double prefix).
- `bun test packages/llm-providers/` — 31 pass, 0 fail.
- `bun run typecheck` (package) — clean.
- Prettier — clean.

Fixes #977

### DIFF
--- a/packages/llm-providers/src/openai-compatible/chat/convert-to-openai-compatible-chat-messages.test.ts
+++ b/packages/llm-providers/src/openai-compatible/chat/convert-to-openai-compatible-chat-messages.test.ts
@@ -74,6 +74,33 @@ describe('user messages', () => {
     ])
   })
 
+  it('should not double-wrap an already-complete data URL', async () => {
+    const result = convertToOpenAICompatibleChatMessages([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'file',
+            data: 'data:image/png;base64,AAECAw==',
+            mediaType: 'image/png',
+          },
+        ],
+      },
+    ])
+
+    expect(result).toEqual([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'image_url',
+            image_url: { url: 'data:image/png;base64,AAECAw==' },
+          },
+        ],
+      },
+    ])
+  })
+
   it('should handle URL-based images', async () => {
     const result = convertToOpenAICompatibleChatMessages([
       {
--- a/packages/llm-providers/src/openai-compatible/chat/convert-to-openai-compatible-chat-messages.ts
+++ b/packages/llm-providers/src/openai-compatible/chat/convert-to-openai-compatible-chat-messages.ts
@@ -41,14 +41,21 @@ export function convertToOpenAICompatibleChatMessages(
                   const mediaType =
                     part.mediaType === 'image/*' ? 'image/jpeg' : part.mediaType
 
+                  const dataUrl =
+                    part.data instanceof URL
+                      ? part.data.toString()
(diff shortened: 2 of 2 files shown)


jahooma:
Thanks for the report and patch. We traced the actual Desktop path end to end. preview_screenshot returns raw base64, not a complete data URL; the corruption happened later because AI SDK 7 adapts file data to a tagged object and our v2 OpenAI-compatible converter serialized that wrapper as [object Object]. The fix is now merged at the provider boundary: tagged inline data and URL values are unwrapped before serialization, while raw v2 inputs remain supported. A full screenshot-history-to-provider-request assertion now produces one valid data:image/png;base64 URL with byte-for-byte matching decoded PNG data. Closing this PR as superseded; the fix will be included in the next Desktop release.

PR #1012, opened 2026-08-14
test(common): add unit test suites for LRUCache and MinHeap

kranthik10:
Adds complete test suites for LRUCache (used in LLM token caching) and MinHeap (priority queue data structure) with 100% test pass rate.


codebuff-team:
Nice contribution. Both suites are well-structured and cover the meaningful edge cases: constructor validation (`common/src/util/__tests__/lru-cache.test.ts` zero/negative maxSize), recency-order eviction behavior on both `get()` and `set()` update, capacity-1 edge case, and generic key/value typing. The MinHeap suite covers empty-heap behavior, ascending extraction order, negative/floating-point scores, duplicate-score ties (correctly asserted via sorted comparison rather than assuming insertion order), and interleaved insert/extract sequences.

This is exactly the kind of test-only PR that's low-risk to port by hand - no production code touched, assertions are specific rather than snapshot-style, and the tests exercise real behavior rather than re-asserting implementation details.

A couple of minor things a maintainer might want before/while porting:
- Confirm the exact error message text in `lru-cache.ts` matches `'LRUCache maxSize must be a positive number.'` exactly, since a typo there would make those two throw-tests fail.
- Consider also testing what happens when `maxSize` is a non-integer (e.g. 1.5), if the implementation guards against that at all - not blocking, just a possible gap.
- No changes to package scripts/CI config were included; assuming `bun test` already picks up `__tests__` directories under `common/src/util`, this should run out of the box.

Overall: good scope, correct layer (tests only, no production code changed), reasonable coverage. Worth porting.

victorxheng:

PR #1026, opened 2026-08-16
fix(sdk): skip plugin code during agent discovery

### LINKED ISSUE #1018: CLI exits after ~6-10s in interactive TUI; bare `freebuff` prints "Usage: node mcp_proxy.js" (Crostini/Debian 12, 0.0.148+0.0.149)
## Summary

On ChromeOS Crostini (Debian 12, x86_64), the freebuff CLI has two reproducible problems:

1. **`freebuff` with zero arguments** prints `Usage: node mcp_proxy.js <serverId_or_socketPath>` and exits 1, instead of starting the CLI.
2. **Any interactive TUI session exits by itself after ~6–10 seconds**, with exit code 1, no error on screen, and nothing on stderr.

Non-TUI commands work perfectly: `--version`, `--help`, and `login` all behave correctly, and `login` waits indefinitely as expected. The fault appears confined to the interactive session path.

Both reproduce on **0.0.148 and 0.0.149** (0.0.149 is current `latest` on npm), so this is not a fresh regression.

## Environment

- OS: Debian GNU/Linux 12 (bookworm) inside ChromeOS Crostini
- Kernel: `6.6.135-09383-g1140e4f27e24`, x86_64
- Terminal: reproduced under a plain pipe, a `script(1)` PTY, and **tmux 3.3a** with `TERM=xterm-256color`
- Install: `npm install -g freebuff` (wrapper `0.0.149`, binary `linux-x64` `0.0.149`)
- Account: authenticated successfully (`freebuff login` completed; the previous 401s disappeared from the logs and ads render)

## Issue 1 — zero-argument invocation

```
$ freebuff
Usage: node mcp_proxy.js <serverId_or_socketPath>
$ echo $?
1
```

Passing **any** argument avoids it:

| Invocation | Result |
|---|---|
| `freebuff` | `Usage: node mcp_proxy.js …`, exit 1 |
| `freebuff --cwd .` | TUI starts |
| `freebuff --continue` | TUI starts |
| `freebuff --help` / `--version` / `login` | correct |

`strace` shows the freebuff process itself writes that line and exits — there is no `node` child and no `mcp_proxy.js` process spawned:

```
3341  write(2, "\33[0m\33[31mUsage: node mcp_proxy.js <serverId_or_socketPath>\33[0m\n", 63)
3341  exit_group(1)
```

The string `serverId_or_socketPath` does **not** appear in the freebuff binary (verified with `strings` across encodings, while other known CLI strings are found), which suggests an external module is being loaded and executed in-process.

## Issue 2 — interactive session exits after ~6–10s

The TUI renders correctly (logo, project picker, and — inside a project — the normal UI with ads), responds to input (arrow keys/Enter navigate the picker), and then the process exits on its own after roughly 6–10 seconds.

The shutdown is **deliberate and graceful**, not a crash: it writes its own terminal-reset sequence, sleeps, unmaps its heap, then `exit_group(1)`. No signal, no core dump, no stderr output.

```
31776 pwritev(1, [{iov_base="\33[?25h\33[0m\33]22;\7\33[>4;0m\33[?1003l…\33[?1049l…"}])
31776 clock_nanosleep(...)
31776 openat(AT_FDCWD, "/dev/tty", O_WRONLY) = -1 ENXIO
31776 exit_group(1)
```

### Possibly relevant: an MCP socket connect immediately precedes shutdown

About 2 ms before the shutdown sequence begins, the freebuff process attempts a UNIX-socket connect that fails:

```
31776 socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0) = 18
31776 connect(18, {sa_family=AF_UNIX, sun_path="/tmp/datacloud-mcp---cwd.sock"}, 110) = -1 ENOENT
31776 close(18)
→ SIGWINCH/SIGQUIT/SIGABRT/SIGPIPE/SIGBUS reset to SIG_DFL
→ terminal reset → exit_group(1)
```

Note the socket name: `datacloud-mcp-` **`--cwd`** `.sock`. That third-party proxy builds its path as `` `/tmp/datacloud-mcp-${idOrPath}.sock` `` where `idOrPath = process.argv[2]`. Here it has picked up **freebuff's own `argv[2]`** (`--cwd`), which implies that proxy module is being loaded into the freebuff process and reading the host's argv.

That single mechanism would explain both issues: with no arguments `argv[2]` is `undefined`, so the module prints its usage banner and calls `process.exit(1)`, killing the host process immediately.

The freebuff binary contains ~28 references to `.agents` and ~25 to `mcpServers`, so it appears to discover agent/MCP plugins from `~/.agents`. This machine has Google's `data-agent-kit-starter-pack` installed there (and under `~/.gemini/extensions`), which ships `mcp/bin/mcp_proxy_bundle.cjs` containing exactly the `Usage: node mcp_proxy.js <serverId_or_socketPath>` string.

Caveat, so this isn't over-claimed: removing that plugin's `.mcp.json`, and moving the entire plugin directory aside, did **not** stop the ~6–10s exit — so either freebuff finds an equivalent definition elsewhere, or the socket connect is a symptom rather than the trigger.

## Ruled out

Each of these was tested and did **not** change the ~6–10s exit:

| Variable | Values tested |
|---|---|
| Credentials | invalid token / no credentials file / freshly authenticated |
| Working directory | `$HOME` / inside a git repo / `/tmp` |
| Terminal | pipe / `script` PTY / tmux, `TERM=xterm-256color` |
| Version | 0.0.148 and 0.0.149 |
| `CODEBUFF_NO_TERMINAL_WATCHDOG=1` | no effect (confirmed `"1"` is an accepted value) |
| Launcher | npm wrapper vs. running the downloaded binary directly |
| Stale config | `settings.json` (`{"mode":"MAX"}`) removed |
| MCP plugin | plugin `.mcp.json` removed; entire plugin directory moved aside |
| Env | `DATA_CLOUD_CURR_IDE_NAME` unset |
| Verbosity | `CODEBUFF_DEBUG=1`, `VERBOSE=1`, `CODEBUFF_TRACE=1` produced no additional output |

The session log (`~/.config/manicode/projects/*/chats/*/log.jsonl`) for a failing authenticated run contains a single benign line (`Enhanced CLI fingerprint generated successfully`) — no error is recorded.

## Additional note (security)

The CLI writes the account API key in **plaintext** into `~/.config/manicode/projects/*/chats/*/log.jsonl`:

```json
{"level":"ERROR","data":{"apiKey":"<redacted>","fields":["id","email"],"status":401}, ...}
```

Worth redacting, since these logs persist on disk and are easy to share when reporting bugs.

## What would help

- Whether the interactive session path is expected to work when unrelated `~/.agents` plugins are present
- A way to start the CLI with MCP/agent-plugin discovery disabled, to confirm or eliminate that path
- A visible error (or a log line) on this exit instead of a silent `exit(1)`

lifelmy:
## Summary

- exclude `.agents/plugins` trees from local agent-module discovery
- add a regression test proving a nested MCP CommonJS executable is never imported

## Root cause

`loadLocalAgents` recursively treated every supported JavaScript or TypeScript file under `.agents` as a candidate agent module and imported it before checking whether it exported an agent definition. Third-party plugin and MCP executables can inspect the host process arguments or call `process.exit()`, so discovery could terminate the Freebuff CLI.

Closes #1018.

## Validation

- `bun test sdk/src/__tests__/load-agents.test.ts --test-name-pattern 'does not execute nested MCP implementation files'`
- `bun test sdk/src/__tests__/load-agents.test.ts` (34 passed)
- `bun run --cwd sdk typecheck`
- `bun run build:sdk`
- `git diff --check`


codebuff-team:
Good, minimal fix that matches the pattern already used for `node_modules`, `scripts`, and `skills*` in `shouldSkipAgentDirectory` (sdk/src/agents/load-agents.ts:111-116). The added test in load-agents.test.ts is a solid regression check — it writes a marker file from inside a nested `plugins/example/mcp/bin/mcp-proxy.cjs` and asserts the marker never appears, which is a much stronger assertion than just checking the export wasn't picked up.

One thing worth flagging for the maintainer doing the port: this closes the door specifically on files under a directory literally named `plugins`, but the underlying problem — `loadLocalAgents` `require`/`import`-ing every `.ts/.js/.cjs` file under `.agents` before checking whether it exports an agent definition — is more general. Any other subdirectory containing a JS file with top-level side effects (not just plugin/MCP bins) would still execute on discovery. That's plausibly a follow-up rather than a blocker for this PR, since it directly addresses the reported issue (#1018) and the plugins/MCP case is the concrete exploit path, but it's not a complete fix for the discovery-model risk.

Scope-wise this is entirely in sdk/src, so no forbidden-path issues. The change is small, tested, and matches existing code conventions, so it's a reasonable candidate to port as-is, with a note that broader hardening of `loadLocalAgents` (e.g., detect exports without executing side effects, or sandbox the require) may be worth a separate issue.

lifelmy:
Thanks for the thorough review. Agreed on both points:

- **Scope of this PR:** it deliberately closes the concrete exploit path from #1018 — a JS/CJS file under a `plugins` directory (the MCP proxy bin) being `require`d during discovery — mirroring the existing `node_modules` / `scripts` / `skills*` skips in `shouldSkipAgentDirectory`. The regression test asserts the side-effect marker never fires, so it fails on the pre-fix loader.
- **Broader discovery-model risk:** you're right that `loadLocalAgents` importing every `.ts/.js/.cjs` under `.agents` before checking for an agent export is the general hazard, and any other subdir with top-level side effects would still execute. That's a larger change (detect exports without executing side effects, or sandbox the require), so I kept it out of this fix to stay minimal and reviewable. Happy to file a separate issue tracking the general hardening if you'd like — just say the word and I'll open it.

Let me know if you'd prefer any adjustment before the port.

lifelmy:
Refreshed this branch onto the latest public `main`; new head is `ad150c4d6`. The merge completed cleanly and only incorporated upstream public snapshot updates. The PR diff remains scoped to the plugin-directory skip in agent discovery and its regression test.

Verification on `ad150c4d6`:
- `git diff --check` -> clean
- `bun test sdk/src/__tests__/load-agents.test.ts` -> 34 pass / 0 fail
- `bun run --cwd sdk typecheck` -> exit 0
- `bun run build:sdk` -> build complete
- `bun run build:freebuff` -> build complete

I use a coding assistant to help implement, and I review and take responsibility for the final change.

lifelmy:
Updated once more after public `main` advanced again; new head is `2158bfe2e`. The merge completed cleanly and only incorporated the latest upstream public snapshot changes. The PR diff remains scoped to the plugin-directory skip in agent discovery and its regression test.

Verification on `2158bfe2e`:
- `git diff --check` -> clean
- `bun test sdk/src/__tests__/load-agents.test.ts` -> 34 pass / 0 fail
- `bun run --cwd sdk typecheck` -> exit 0
- `bun run build:sdk` -> build complete
- `bun run build:freebuff` -> build complete

I use a coding assistant to help implement, and I review and take responsibility for the final change.

lifelmy:
Refreshed this branch onto the latest public `main`; new head is `7cee43529`. The merge completed without manual conflicts and only incorporated upstream public snapshot updates (`bun.lock`, `common/src/constants/freebuff-ads.ts`).

Verification run after the refresh:
- `git diff --check`
- `bun test sdk/src/__tests__/load-agents.test.ts` (34 pass / 0 fail)
- `bun run --cwd sdk typecheck`
- `bun run build:sdk`
- `bun run build:freebuff`

I use a coding assistant to help implement, and I review and take responsibility for the final change.

lifelmy:
Refreshed this branch onto current public `main`; new head is `d82aa69ca`. The merge completed without manual conflicts and only brought in the upstream `common/src/types/freebuff-session.ts` change.

Verified after refresh:
- `git diff --check`
- `bun test sdk/src/__tests__/load-agents.test.ts` (34 pass / 0 fail / 69 expects)
- `bun run --cwd sdk typecheck`
- `bun run build:sdk`
- `bun run build:freebuff`

I use a coding assistant to help implement, and I review and take responsibility for the final change.

lifelmy:
Refreshed this branch onto the latest public `main`; new head is `f95d144e6`. The merge completed without manual conflicts and only incorporated the latest upstream public snapshot updates (`README*`, free-model catalog/constants/tests, and related CLI settings coverage). The PR diff remains scoped to skipping plugin code during agent discovery and its regression test.

Verification after refresh:
- `bun test sdk/src/__tests__/load-agents.test.ts` -> 34 pass / 0 fail
- `bun run --cwd sdk typecheck` -> exit 0
- `bun run build:sdk` -> success
- `bun run build:freebuff` -> success

I use a coding assistant to help implement, and I review and take responsibility for the final change.

victorxheng:

PR #1083, opened 2026-08-22
fix(cli): terminal detection for WezTerm/Ghostty/Warp/Konsole + Windo…

smarpitm:
…ws pwsh fallback

Add kitty graphics protocol detection for WezTerm, Ghostty, Warp, and Konsole terminals (case-insensitive TERM_PROGRAM matching + KONSOLE_VERSION).

Fix Windows clipboard image paste failing silently when powershell.exe is missing by falling back to pwsh (PowerShell 7). Normalize stdout/stderr to strings so callers don't handle the string | Buffer union.

Also includes spec-compliant kitty chunking (m=0 terminator on last chunk, control data only on first chunk), iTerm2 size param fix (decoded bytes not base64 length), and image pipeline integrity tests.

Split from a912ea6 — the getKittyFormat() format ID fix is deferred to a follow-up PR.

🤖 Generated with Codebuff


codebuff-team:
Good work here — several of these are genuine, well-reasoned bug fixes rather than cosmetic churn:

- `generateKittyImageSequence` in `cli/src/utils/terminal-images.ts`: the old code repeated the full control block (`a=T,f=100,...`) on every chunk and never emitted `m=0` on the final chunk, which is a real spec violation that could leave transmissions unterminated on strict kitty implementations. The rewrite (only first chunk gets full control data, subsequent chunks carry `m=` only, last chunk gets `m=0`) is correct and matches the kitty graphics protocol doc you linked. The new `terminal-images.test.ts` chunk-boundary assertions back this up well.
- `generateITerm2ImageSequence`'s `size=` param bug (base64 length vs decoded byte length) was a real, findable-in-review defect; the fix and the accompanying test (`size=5` not `size=8` for `'aGVsbG8='`) are both correct.
- The WezTerm/Ghostty/Warp/Konsole detection additions are low-risk, env-var-only, and tested.
- The `spawnPowerShell` fallback to `pwsh` in `clipboard-image.ts` is a reasonable idea for minimal Windows installs, but it's untested — there's no test exercising the ENOENT-then-pwsh path, and the silent timeout bump in `readClipboardFilePathWindows` (1000ms → 5000ms) isn't mentioned in the PR description, so a reviewer can't tell if that's intentional or incidental.

My main critique is scope: this bundles three fairly independent changes (protocol chunking fix, terminal detection, Windows shell fallback) into one PR. Splitting the clipboard/pwsh piece out with its own test would make it easier to port and verify in isolation. The `test/setup-scm-loader.ts` stub is a sensible, clearly-commented mirror-only shim and shouldn't be ported as-is.

Overall the core protocol fixes are worth taking; the Windows fallback piece would benefit from a test before landing.

victorxheng:

PR #1092, opened 2026-08-23
fix: handle Windows startup access violations

### LINKED ISSUE #1091: freebuff.exe crashes immediately with access violation (0xC0000005) on Windows x64
## Description

Image

`freebuff` crashes immediately when launched on Windows.

The process exits with code `3221225477` and reports an access violation.

### Environment

- Platform: win32 x64
- Node: v24.15.0
- Target: win32-x64
- Binary: `C:\Users\8292a\.config\manicode\freebuff.exe`

### Error

```text
freebuff exited immediately (code 3221225477)

The binary crashed with an access violation.

### Steps to reproduce

1. Install and launch Freebuff on a Windows x64 system.
2. Open a terminal such as PowerShell.
3. Change to another directory, for example:
   `cd F:\`
4. Run `freebuff`.
5. Observe that `freebuff` exits immediately with code `3221225477` and reports an access violation.

### Where does this happen?

CLI (terminal client)

### Operating system

Windows

### Version

0.0.149

### Model

_No response_

### Logs or screenshots

```shell

```

abhiprd2000:
## Summary

Fixes #1091.

On Windows, a `0xC0000005` (`STATUS_ACCESS_VIOLATION`) crash from the optimized binary during the startup window can indicate a CPU compatibility problem. The launcher already falls back to the baseline binary for `0xC0000409`, but did not handle `0xC0000005`.

This change:

- treats `0xC0000005` as a startup CPU-feature crash
- keeps the existing startup-time guard so normal runtime access violations are unaffected
- adds regression coverage for the new fallback case

### DIFF
--- a/cli/release-core/launcher.js
+++ b/cli/release-core/launcher.js
@@ -127,9 +127,10 @@ function createLauncher(productConfig) {
    * perfectly capable machine to the slower build.
    */
   function isStartupCpuFeatureCrash(code, signal, msAlive) {
+    const unsignedCode = getUnsignedExitCode(code)
     return (
       process.platform === 'win32' &&
-      getUnsignedExitCode(code) === 0xc0000409 &&
+      (unsignedCode === 0xc0000409 || unsignedCode === 0xc0000005) &&
       !signal &&
       typeof msAlive === 'number' &&
       msAlive < STARTUP_CRASH_WINDOW_MS
--- a/cli/src/__tests__/launcher-avx2-fallback.test.ts
+++ b/cli/src/__tests__/launcher-avx2-fallback.test.ts
@@ -340,9 +340,10 @@ describe('illegal-instruction detection', () => {
  * STATUS_STACK_BUFFER_OVERRUN instead — and those users crash-looped forever.
  */
 describe('windows startup-abort detection', () => {
-  test('recognizes an immediate 0xC0000409 as a CPU-feature suspect', () => {
+  test('recognizes immediate Windows startup CPU-feature crashes', () => {
     const t = makeLauncher()
     expect(t.isStartupCpuFeatureCrash(0xc0000409, null, 40)).toBe(true)
+    expect(t.isStartupCpuFeatureCrash(0xc0000005, null, 40)).toBe(true)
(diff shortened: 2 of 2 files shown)


codebuff-team:
Thanks for looking into the Windows startup-crash handling — the `isStartupCpuFeatureCrash` guard and the accompanying test update are the right places to make this change, and the diff is small and readable.

The concern is with the substance, not the mechanics: `0xC0000005` (`STATUS_ACCESS_VIOLATION`) is an extremely generic NTSTATUS code. Unlike `0xC0000409` (stack buffer overrun, which reliably correlates with illegal-instruction crashes from AVX2/AVX-512 misuse), access violations can come from null derefs, use-after-free, heap corruption, bad DLL loads, etc. — bugs that have nothing to do with CPU feature mismatch. Silently falling back to the baseline binary on *any* access violation in the startup window risks hiding genuine crash bugs in the optimized build rather than surfacing them, which is a real regression in diagnosability even if it stops one class of crash-loop.

The old test explicitly asserted `0xc0000005` was *not* this kind of crash (`other native crashes and signals are not this one`), and this PR just flips that assertion rather than narrowing the detection. That suggests the original exclusion may have been deliberate, not an oversight — it'd help to know what evidence from #1091 shows this is specifically a CPU-feature issue and not a real access violation in the optimized binary. Some kind of secondary signal (e.g. only treat as CPU-feature-crash if immediate on process start with 0ms alive, or a targeted allowlist of instruction pointers) would make this safer than a blanket code match.

Worth reopening with more evidence from the linked issue before this masks a legitimate crash for someone else.

abhiprd2000:
@codebuff-team Ok, you were right!

The machine is an i5-5200U (Broadwell), which has AVX2, so the CPU-feature theory was wrong from the start.

More importantly, the Application log shows chrome.exe, Code.exe, updater.exe and one process Windows could not even name all crashing with 0xC0000005 in the same four minute window, every one with "Faulting module name: unknown",
and every one with a fault offset of 0x...44d0.

So 0xC0000005 was exactly as generic as you said. Closing.

PR #1101, opened 2026-08-23
fix(common): use product-specific commit attribution

### LINKED ISSUE #997: Generated By Codebuff when doing a pull request.
Image

Not sure if this is intended but it still uses CodeBuff when doing pull requests.

c8dhjp4tyv-bit:
## Summary
- derive git commit attribution from the compile-time `FREEBUFF_MODE` flag
- keep Codebuff's existing footer for standard builds
- use Freebuff branding and domain for Freebuff binaries
- add regression coverage for both product variants

Fixes #997

## Validation
- Freebuff build-flag attribution smoke check: passed
- targeted common tool tests: 10 passed
- common typecheck: passed
- `git diff --check`: passed
- Prettier check for changed files: passed
- common full test command: 1,158 passed; 37 existing checkout/snapshot failures remain in unrelated web/agent sources and one existing Zod schema expectation
- agent-runtime typecheck: existing missing `agents-graveyard/researcher/researcher` imports; no errors from changed files


codebuff-team:
Good instinct and clean fix for #997: Freebuff binaries were asking the model to sign commits as Codebuff, which is wrong for a product with its own identity. The new `getGitCommitGuidePrompt(isFreebuff)` function is testable and the added test file (`run-terminal-command-branding.test.ts`) covers both variants directly, which is exactly what this repo would want for a behavior change like this.

One thing worth cleaning up before this lands: `commitProductName`/`commitProductDomain` (lines ~136-137) duplicate the exact same ternary logic that's already inside `getGitCommitGuidePrompt`. Since that function takes an optional `isFreebuff` argument, you could just call `getGitCommitGuidePrompt(isFreebuffBuild)` once and derive the example string from a shared value, or export the `productName`/`productDomain` pair from the function so there's a single source of truth. Right now if someone touches the Freebuff domain/name in one place they can easily miss the other.

Also worth a comment or test noting that `isFreebuffBuild` is read from `process.env.FREEBUFF_MODE` at module load time (not truly compile-time despite the comment) — if any test or runtime code mutates that env var after this module is first imported, `gitCommitGuidePrompt` (the pre-computed export) won't reflect it. Not a blocker, but worth flagging since the PR description calls it a compile-time flag.

Overall this is a small, well-scoped, well-tested fix and worth porting after tightening the duplication.

c8dhjp4tyv-bit:
Addressed both cleanup points in 12835624ec287b0c4b6356dfbf6c7bf088396f3f.

- Product name/domain now come from one `getCommitAttribution()` helper, used by both `getGitCommitGuidePrompt()` and the native tool-call example. The duplicated Freebuff/Codebuff ternaries are gone.
- Clarified the `FREEBUFF_MODE` lifecycle in the source comment: it is injected for Freebuff builds and read once at module initialization. The tests intentionally pass an explicit product flag, so they do not depend on mutating `process.env` after the module has already been imported.

victorxheng:

PR #1102, opened 2026-08-23
fix(cli): disable Freebuff chat logo sheen

### LINKED ISSUE #1035: Can the FREEBUFF flashing effect displayed by CLI be turned off? It is currently causing the input method to flicker
Image

Can the FREEBUFF flashing effect displayed by CLI be turned off? It is currently causing the input method to flicker

c8dhjp4tyv-bit:
## Summary
- disable the animated chat-header logo sheen in Freebuff builds
- keep the existing animation for standard Codebuff builds
- add focused regression coverage for both product paths

Fixes #1035

## Validation
- focused animation tests: 3 passed
- related queue tests: 4 passed
- full CLI suite: 2,367 passed, 9 skipped; existing checkout/environment failures remain (34 failed tests and 32 harness errors)
- CLI typecheck: existing missing `tar` and `react-dom/server` declarations; no errors from changed files
- Prettier check for changed files: passed
- `git diff --check`: passed


codebuff-team:
Nice, minimal fix. The extraction of `shouldAnimateChatHeader` into its own pure function in `cli/src/utils/chat-header-animation.ts` is the right call — it's independently testable and keeps `chat-header.tsx` unchanged in structure (single line diff). Tests in `chat-header-animation.test.ts` cover the three meaningful branches (Freebuff-off, Codebuff-on, and inactive-in-either-case), which is proportionate to the size of the change.

A couple of things a maintainer will likely want confirmed before porting:
1. Confirm `IS_FREEBUFF` is genuinely a build-time/product flag already used elsewhere in `cli/src/utils/constants.ts` — the diff assumes it exists and is already threaded through `chat-header.tsx`, which it is per the import list, so this looks consistent with existing conventions rather than introducing a new product branch.
2. The linked issue (#1035) presumably describes the flicker/visual bug — worth double-checking the fix is actually disabling the *right* animation and not just papering over a timing bug in `useSheenAnimation` itself that could resurface elsewhere.

Overall this is a clean, scoped, well-tested change that should port cleanly by hand.

c8dhjp4tyv-bit:
Verified both points against the current tree:

1. `IS_FREEBUFF` is the existing Freebuff build-time product flag in `cli/src/utils/constants.ts`; it is injected via `--define`/`FREEBUFF_MODE` and is already the flag imported by `chat-header.tsx`.
2. Issue #1035 is specifically about the flashing Freebuff **logo** causing input-method flicker, and its discussion explicitly asks to turn off the LOGO animation. This PR disables only the chat-header sheen for Freebuff while leaving Codebuff and inactive-header behavior unchanged.

So I don't see a broader `useSheenAnimation` timing bug that this change is masking; the scoped behavior matches the reported request.

victorxheng:

PR #1104, opened 2026-08-23
docs(cli): document account switching

### LINKED ISSUE #1086: change account freebie
### What happened

please documentation  for change account  in `freebuf`

### Steps to reproduce

please documentation  for change account  in `freebuf`

### Where does this happen?

CLI (terminal client)

### Operating system

macOS

### Version

154

### Model

_No response_

### Logs or screenshots

```shell

```

c8dhjp4tyv-bit:
## Summary

- document how to switch the account used by the CLI
- mention the existing `/logout` and `/signout` commands
- point users to the `freebuff login` browser flow after logout

Fixes #1086

## Validation

- verified `command-registry.ts` defines `/logout` with `/signout` as an alias
- verified `login-modal.tsx` prompts `Press ENTER to login...` before starting the browser login flow
- verified `cli-args.ts` accepts `freebuff login`
- `git diff --check` passed


codebuff-team:
Nice, small, focused doc addition that fills a real gap (issue #1086) — README currently doesn't mention how to switch accounts. The wording is clear and the placement right after the intro flow makes sense.

Before this gets ported, please double check accuracy against actual CLI behavior:

- Confirm both `/logout` and `/signout` are real, currently-supported commands (not just aliases in some older version). If only one exists, drop the other to avoid confusing users.
- Confirm "press Enter on the login screen" is accurate — if the login screen auto-launches the browser flow without needing Enter, or if the flow differs, this could mislead readers.
- The "Validation" section in the PR description mentions CLI argument/alias tests, but this is a docs-only diff — it's unclear those tests actually exercise this claim. Worth clarifying what was actually verified, or removing that section if it's not directly applicable.

Assuming the described behavior is verified against the current CLI, this is a good small contribution worth porting.

c8dhjp4tyv-bit:
Verified the README wording directly against the current CLI behavior:

- `command-registry.ts` defines `/logout` with `aliases: ['signout']`, so both `/logout` and `/signout` are current commands.
- `login-modal.tsx` renders `Press ENTER to login...` before opening the browser flow, so the Enter instruction is accurate.
- `cli-args.ts` explicitly allows `login` as the Freebuff CLI command, so `freebuff login` is supported.

The original validation line was broader than needed for this docs-only change; the three checks above are the direct implementation evidence for every behavior documented by the new paragraph.

victorxheng:

PR #1120, opened 2026-08-25
feat(cli): add /doctor command for environment and dependency diagnostics

### LINKED ISSUE #1119: Add /doctor command to diagnose local environment and dependencies
When running the CLI, it can be hard to tell why something fails if local tools or dependencies aren't set up properly.

For example:
- Code search failing when ripgrep (`rg`) isn't found or executable
- Git tracking issues when `git` isn't in PATH or the current folder isn't a repo
- Connectivity/firewall issues reaching the API
- Missing `tmux` for multiplexing/detached sessions

Currently, `/diagnostics` only shows process memory/CPU and PID watchdog info, so there isn't a quick way to check if local dependencies and tooling are working properly.

nilaymallikk:
### Summary

Closes #1119.

Adds a `/doctor` slash command (with aliases `/health` and `/check`) to the CLI that runs fast local diagnostics to verify that required tools and environment prerequisites are healthy:

- **Runtime & OS**: Reports Bun/Node version and platform architecture.
- **Git**: Validates `git` binary and whether the current directory is an active Git repo (with branch name).
- **Ripgrep (`rg`)**: Resolves the bundled ripgrep executable from `@codebuff/sdk` and verifies it executes properly (with fallback to system `rg`).
- **Tmux**: Verifies `tmux` availability for detached sessions.
- **Network Connectivity**: Fast reachability check to the configured API endpoint.

### Scope

All changes are strictly scoped within `cli/`:
- `cli/src/commands/doctor.ts` (new)
- `cli/src/commands/command-registry.ts`
- `cli/src/data/slash-commands.ts`
- `cli/src/commands/__tests__/doctor.test.ts` (new unit tests)

### Verification

- Passed unit tests: `bun test cli/src/commands/__tests__/doctor.test.ts`
- Passed full CI build and packaging: `bun run ci`
- Passed binary smoke test: `bun cli/scripts/smoke-binary.ts cli/bin/freebuff`


codebuff-team:
Nice feature: `/doctor` (with `/health` and `/check` aliases) gives a clear, useful diagnostic report, and the implementation follows the codebase's existing `collectProcessDiagnostics`/`formatProcessDiagnostics` pattern closely, which makes it easy to review and port. Tests in `doctor.test.ts` cover each check function plus the formatter and command-registry resolution — good coverage for a first PR.

A few things worth double-checking before this lands upstream:

1. `checkGit`, `checkTmux`, and `checkRipgrep` all use `spawnSync` with a 5s timeout each. In the worst case (git missing, tmux missing, rg falling back to system check) that's up to ~15s of blocking calls before the network check even starts. Consider running these in parallel with `Promise.all` around async wrappers, or trimming the timeouts, since users invoking `/doctor` expect a quick response.
2. `checkNetwork`'s fallback URL `'https://freebuff.com'` is branding-specific; given the codebase supports dual branding via `IS_FREEBUFF`, it might be cleaner to derive the fallback from the same env vars used elsewhere rather than hardcoding one brand's domain.
3. Worth confirming `getBundledRgPath` is actually exported from `@codebuff/sdk` in the target repo version — the PR assumes it exists but I can't verify from the diff alone.
4. Minor: unused `getCliEnv`/`env` fallback chain in `checkNetwork` (`env.NEXT_PUBLIC_CODEBUFF_APP_URL || env.NEXT_PUBLIC_CODEBUFF_APP_URL || fallback`) has redundant duplicate reference — likely a copy-paste artifact, should be cleaned up.

None of these are blockers; the core idea and execution are solid and in-scope for `cli/`. Recommend tightening the parallelism and the URL fallback before a maintainer ports this.

victorxheng:

PR #1133, opened 2026-08-26
Add GitHub Actions workflow for Node.js with Webpack

markusknudtsen-droid:


codebuff-team:
Thanks for the contribution, but this appears to be an unmodified GitHub-generated template (the default "Node.js with Webpack" starter workflow) rather than something tailored to this repo.

Codebuff/Freebuff isn't built with webpack - there's no `npx webpack` entry point in this codebase, so this workflow would simply fail on every push/PR to `main`. It also duplicates CI concerns that should live alongside the project's actual build tooling (which is out of scope for this mirror anyway, per packages/build-tools/).

Since this would add a broken, always-failing CI job with no relation to the project's real build process, I can't recommend porting it. If you're interested in improving CI, it'd help to first look at how the project actually builds/tests (e.g. via the existing scripts in package.json) and propose a workflow that matches that, rather than a generic scaffold.

Closing as not applicable to this codebase.

victorxheng:

Issue #1168, opened 2026-08-31
CLI: torn run-state.json silently resumes chats with no agent context (model amnesiac, transcript intact)


nordicnode:
Follow-up from #1166 (the compaction wipes). While auditing that report I traced the persistence side and found a second, independent path to the same symptom — "the model lost all chat history" — with a different cause. Verified against the current code with executable repros.

## The failure chain

1. **The atomic write is not durable.** `cli/src/utils/write-file-atomic.ts` writes a temp file and renames it over `run-state.json` — but never fsyncs. The rename is atomic against concurrent processes, yet after a power loss / hard hang / battery pull, the filesystem can land the rename while the file's data blocks were never written. Result: a truncated or empty `run-state.json` exactly where the atomic rename was supposed to guarantee a complete file. (The in-code comment acknowledges torn *non-atomic* writes; the same tear reaches through this pattern without the fsync.)

2. **The load path has no recovery.** `loadMostRecentChatState` (`cli/src/utils/run-state-storage.ts`) parses `run-state.json` and, on failure, falls back to a placeholder:
   ```ts
   runState ??= { output: { type: 'error', message: 'Previous run state could not be restored.' } } as RunState
   ```
   The placeholder has **no `sessionState`**. Meanwhile `chat-messages.json` is parsed independently and usually survives — so the UI transcript is fully intact.

3. **A missing `sessionState` means a fresh session.** `sdk/src/run.ts:436-467`: when `previousRun.sessionState` is undefined, `runOnce` builds a brand-new `initialSessionState`. The next turn's model context = empty history.

Net effect: the user resumes a chat, sees their whole transcript, sends a message — and the assistant has no idea who they are or what it did. That is precisely the "lost all chat history instead of compacting" report shape, on machines that crashed or lost power mid-session. **Reproduced** with a temp-dir test: a truncated primary + intact transcript yields a placeholder RunState without `sessionState`.

Two aggravating details found on the way:

- **Recoverable state is thrown away.** `writeFileAtomic` leaves `run-state.json.<pid>.<uuid>.tmp` behind when the process dies between write and rename — that temp is usually a *complete* generation of the state. The loader ignores it.
- **The loss is silent in the UI.** Nothing tells the user the assistant's memory is gone; they just experience the model being broken.

## Suggested fix (all verified in my PR)

1. fsync the temp file before the rename in both `writeFileAtomic` and `writeFileAtomicAsync` (rename becomes durable *and* its data with it).
[...]

PR #1169, opened 2026-08-31
fsync atomic writes before rename and retry Windows rename locks

nordicnode:
## Summary

Durability half of #1169, narrowed per review to the root-cause fix: the recovery/rotation/UI machinery now lives in #1195 so each piece can be reviewed and perf-checked separately.

- **`writeFileAtomic`/`writeFileAtomicAsync` now fsync the temp before renaming.** Without it, a power loss can land the rename while the data blocks were never written — a torn `run-state.json` exactly where the atomic rename was supposed to guarantee a complete file, which is what made resumed chats amnesiac (details in #1168). EINVAL filesystems that reject fsync degrade gracefully to the old rename-only guarantee.
- **The async rename retries briefly on EPERM/EBUSY/EACCES.** On Windows the just-closed handle can still be held by AV/indexer scans — a transient the fsync widens the window for. Bounded (5 attempts, exponential backoff), non-matching errors rethrow immediately. Now covered by tests (previously prose-only, per review).

## Root cause

`rename(2)` is atomic but not durable: it orders the name against the filesystem metadata journal, not the file's data blocks. A rename that survives a power cut can still point at a file whose contents were never flushed, so the "atomic" write yields a torn file. fsync on the temp fd before the rename closes that gap.

## Changes

1. `cli/src/utils/write-file-atomic.ts` — `fsyncFile(fd)` after write, before rename, in both the sync and async writers; `renameWithRetry` in the async path for the Windows handle-release race.

## Tests

- 3 new tests in `cli/src/utils/__tests__/write-file-atomic.test.ts`: fsync-before-rename ordering (spied `fs.fsyncSync`/`fs.renameSync` call order), transient EPERM retry succeeds on attempt 3, non-transient ENOENT rethrows on attempt 1.
- `NODE_ENV=production bun test src/utils/__tests__/write-file-atomic.test.ts` — 13 pass (10 pre-existing + 3), 0 fail.

## Validation

- `NODE_ENV=production bun test src/utils/__tests__/run-state-storage.test.ts src/utils/__tests__/write-file-atomic.test.ts` — 50 pass, 0 fail.
- `bun x tsc --noEmit -p cli/tsconfig.json` — 10 errors, all pre-existing on `main`, none in the changed files.
- `bunx prettier --check` clean on both changed files.

Refs #1168, #1195


nordicnode:
Updated implementation now addresses parent-directory fsync gap per reviewer feedback. New work here: https://github.com/nordicnode/freebuff/pull/1

Summary of changes vs #1169:
- Added `fsyncParentDir()` after rename to protect FRESHNESS (newest generation wins over older after power loss)
- Bench: ~0.0026 ms/op dir cost on tmpfs; full 5 MiB payload ~2.3 ms/op
- Two new ordering tests verify `fsync→rename→fsync` and async equivalent

This is strictly additive to the current diff (no scope creep), fixes the durability gap without changing test counts or file operations count, and answers all four objections raised in the codebuff-team comment.

PR #1249, opened 2026-09-03
Fix NaN handling in min-heap index calculations

pavankumar-vh:
## Overview
Fix NaN handling in min-heap index calculations in `common/src/util/min-heap.ts`.

## Bug Description
The functions didn't validate that index is a finite number. If index was NaN or Infinity, `Math.floor((NaN - 1) / 2)` would return NaN, causing the heap operations to fail.

## Fix
Added `Number.isFinite()` checks to default to 0 for invalid numbers.

## Testing
No existing tests for this function, but the fix prevents runtime errors with invalid inputs.

## Files Changed
- `common/src/util/min-heap.ts` - Added NaN validation

## Scope

### DIFF
--- a/common/src/util/min-heap.ts
+++ b/common/src/util/min-heap.ts
@@ -5,14 +5,17 @@ export class MinHeap<T> {
   private heap: { item: T; score: number }[] = []
 
   private getParentIndex(index: number): number {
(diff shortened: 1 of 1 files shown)


codebuff-team:

PR #1252, opened 2026-09-03
Fix Infinity validation in addDaysToDateKey

pavankumar-vh:
## Overview
Fix Infinity validation in the `addDaysToDateKey` function in `common/src/util/freebuff-streak.ts`.

## Bug Description
The function only checked for NaN, not Infinity. If the date string represented an extreme date, `getTime()` could return Infinity, causing the date arithmetic to produce invalid results.

## Fix
Changed `Number.isNaN()` to `Number.isFinite()` to catch both NaN and Infinity.

## Testing
No existing tests for this function, but the fix prevents incorrect behavior with extreme dates.

## Files Changed
- `common/src/util/freebuff-streak.ts` - Added Infinity validation

## Scope


codebuff-team:

PR #1263, opened 2026-09-03
Add escapeHtml function for HTML context escaping

pavankumar-vh:
## Overview
Add a dedicated `escapeHtml` function for escaping strings in HTML contexts, separate from the existing `escapeString` function.

## Problem
The previous PR (#1231) modified `escapeString` to escape HTML-unsafe characters, but this could break existing callers that depend on the un-escaped characters passing through (e.g., building strings for JSON parsing or non-HTML contexts).

## Fix
Added a separate `escapeHtml` function that explicitly escapes HTML-unsafe characters (`<`, `>`, `&`, `'`) to prevent XSS attacks and HTML injection. The existing `escapeString` function remains unchanged for backward compatibility.

```typescript
export const escapeHtml = (str: string): string => {
  return JSON.stringify(str)
    .slice(1, -1)
    .replace(/</g, '\\u003c')
    .replace(/>/g, '\\u003e')
    .replace(/&/g, '\\u0026')
    .replace(/'/g, '\\u0027')
}
```

## Testing
Added comprehensive test coverage for both functions:
- `escapeString` tests verify it still escapes JSON special characters but NOT HTML-unsafe ones
- `escapeHtml` tests verify it properly escapes `<script>`, `&`, `'`, `>` to prevent XSS
- Both functions tested with empty strings and regular characters

All 24 tests pass (19 existing + 5 new).

## Files Changed
- `common/src/util/string.ts` - Added escapeHtml function
- `common/src/util/__tests__/string.test.ts` - Added test coverage for both functions

## Scope

### DIFF
--- a/common/src/util/__tests__/string.test.ts
+++ b/common/src/util/__tests__/string.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'bun:test'
 
-import { pluralize } from '../string'
+import { escapeHtml, escapeString, pluralize } from '../string'
 
 describe('pluralize', () => {
   it('should handle singular and plural cases correctly', () => {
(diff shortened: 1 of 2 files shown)


pavankumar-vh:
Closing this PR - the escapeHtml function is a dead export with no consumers. The maintainer feedback was correct that this doesn't fix a real bug since no code paths actually use it in HTML contexts.

PR #1264, opened 2026-09-03
Add defensive check for data URL prefix in summarizeDataUrl

pavankumar-vh:
## Overview
Add a defensive check for the data URL prefix in `summarizeDataUrl` function in `common/src/util/cache-debug.ts`.

## Bug Description
The `summarizeDataUrl` function was called from `summarizeLargeValue` which checks if a string starts with `data:` before calling it. However, if `summarizeDataUrl` were ever called from a different context or if the prefix check were removed, the function would incorrectly try to parse non-data-URL strings, potentially producing misleading summaries.

## Fix
Added a defensive check at the start of `summarizeDataUrl` that returns the value unchanged if it doesn't start with `data:`. This makes the function safer to use independently and prevents incorrect parsing of malformed inputs.

## Testing
Added comprehensive test coverage demonstrating:
1. Valid data URLs are properly summarized (image/png with base64 payload)
2. Non-data-URL strings (https://example.com/image.png) pass through unchanged
3. Edge case: data URL-like strings without comma are handled gracefully

All 3 tests pass.

## Files Changed
- `common/src/util/cache-debug.ts` - Added defensive prefix check
- `common/src/util/__tests__/cache-debug.test.ts` - New test file with 3 test cases

## Scope

### DIFF
--- a/common/src/util/__tests__/cache-debug.test.ts
+++ b/common/src/util/__tests__/cache-debug.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from 'bun:test'
+
+import { normalizeProviderRequestBodyForCacheDebug } from '../cache-debug'
+
+describe('cache-debug data URL handling', () => {
+  it('summarizes valid data URLs', () => {
+    const result = normalizeProviderRequestBodyForCacheDebug({
+      provider: 'openai',
+      body: {
+        model: 'gpt-4',
+        messages: [
+          {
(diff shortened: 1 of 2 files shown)


pavankumar-vh:
Closing this PR - the defensive check in summarizeDataUrl is dead code since the function is only called from one place that already validates the input. The maintainer feedback was correct that this is speculative hardening without a real bug being fixed.

PR #1271, opened 2026-09-04
Fix NaN handling in ttftBucketIndex

pavankumar-vh:
## Overview
Fix NaN handling in the `ttftBucketIndex` function in `common/src/util/ttft-histogram.ts`.

## Bug Description
The function didn't validate that ttftMs is a finite number. If ttftMs was NaN or Infinity, `Math.max(NaN, 1)` would return NaN, causing `Math.log(NaN)` to return NaN, and the entire calculation would produce NaN.

## Fix
Added `Number.isFinite()` check to default to 0 for invalid numbers.

## Testing
The existing comprehensive test suite (11 tests, 995 assertions) already covers edge cases including NaN, Infinity, negative values, monotonicity, and accuracy bounds. All tests pass with this fix.

The test suite includes:
- Monotonicity checks across latency ranges
- Degenerate sample handling (0, 0.4, 1)
- Pathological sample clamping (MAX_SAFE_INTEGER)
- Relative error bounds (< 1.6%)
- Percentile accuracy on realistic distributions

## Files Changed
- `common/src/util/ttft-histogram.ts` - Added NaN/Infinity validation

## Scope

### DIFF
--- a/common/src/util/ttft-histogram.ts
+++ b/common/src/util/ttft-histogram.ts
@@ -37,7 +37,8 @@ const LN_BASE = Math.log(TTFT_HISTOGRAM_BASE)
  * Sub-millisecond and zero samples land in bucket 0 rather than at -Infinity.
  */
 export function ttftBucketIndex(ttftMs: number): number {
-  const index = Math.floor(Math.log(Math.max(ttftMs, 1)) / LN_BASE)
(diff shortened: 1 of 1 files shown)


pavankumar-vh:
Closing - duplicate of PR #1284 which has the correct fix.

PR #1346, opened 2026-09-13
feat(cli): add -c/--continue alias and 'nothing to continue' notice

### LINKED ISSUE #1338: Add "freebuff -c" / "freebuff --continue" to resume an active session
## Problem

Sometimes the Freebuff CLI closes unexpectedly while a 1-hour session is still active, for example because of a crash, terminal closing, or an update.

## Proposed solution

It would be really useful to have a way to continue that session instead of wasting the remaining time.

Both of these should work:

```bash
freebuff -c
freebuff --continue
```

They should resume the most recently active session if it hasn't expired or been explicitly ended, including its existing conversation/state and remaining session time.

> If there is no active session to resume, Freebuff should clearly tell the user that there is nothing to continue.

## Benefit

This would prevent users from losing the rest of their session just because the CLI happened to close or restart.

shauryagangrade:
## Summary

Implements part of the ask in #1338: `freebuff -c` (and the long form `freebuff --continue`) now resume an active session instead of wasting the remaining time when the CLI crashed, the terminal closed, or an update interrupted the previous run.

### Changes

- **`cli/src/cli-args.ts`** — added the `-c` short alias to `--continue [conversation-id]` (freebuff and codebuff parsers).
- **`cli/src/index.tsx` / `cli/src/app.tsx`** — thread the raw `continueRequested` flag (distinct from history-pick resume) down to the landing screen.
- **`cli/src/components/freebuff-landing-screen.tsx`** — when `--continue` was requested but the session probe finds no active seat to resume (`none`, expired, or explicitly ended), the landing screen now clearly says there is nothing to continue. Rows are accounted for in the picker's height budget.
- Tests for flag parsing (`cli-args.test.ts`) and the notice render (`freebuff-nothing-to-continue.test.tsx`).

### Resume behavior

The actual session resume relies on the existing free session infra: on startup the probe auto-takes over the server seat when the prior local process is dead (the crash/terminal-close/update case), which preserves the server session and its remaining time, while `--continue` restores the previous conversation state. When the seat is held by a live process, the existing "Take over" confirmation is shown rather than silently superseding it.

### Verification

- New + adjacent tests pass; full `cli` suite has no new failures (the 21 failures/19 errors are pre-existing at HEAD: missing private `packages/internal` env + release-wrapper checks).
- Typecheck clean for the touched files (only pre-existing `tar`/`react-dom/server` declaration errors remain).

Closes #1338

### DIFF
--- a/cli/src/__tests__/cli-args.test.ts
+++ b/cli/src/__tests__/cli-args.test.ts
@@ -178,4 +178,37 @@ describe('Freebuff CLI Argument Parsing', () => {
     expect(result.command).toBe('login')
     expect(result.initialPrompt).toBeNull()
   })
+
+  test('accepts -c as continue without a conversation id', () => {
+    const result = parseArgs({
+      argv: ['node', 'freebuff', '-c'],
+      isFreebuff: true,
+      version: '1.0.0',
+    })
+
+    expect(result.continue).toBe(true)
+    expect(result.continueId).toBeNull()
+  })
+
+  test('accepts --continue as continue without a conversation id', () => {
+    const result = parseArgs({
+      argv: ['node', 'freebuff', '--continue'],
+      isFreebuff: true,
+      version: '1.0.0',
+    })
+
+    expect(result.continue).toBe(true)
+    expect(result.continueId).toBeNull()
+  })
+
+  test('accepts -c with an explicit conversation id', () => {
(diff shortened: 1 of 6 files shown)


shauryagangrade:
@codebuff-team review

PR #1357, opened 2026-09-15
fix(cli): keep the session readout in its own lane so a transient status cannot blank it

KazenDev:
The status bar draws one thing per row. Its idle label — model, time left, context occupancy — is the **last** entry in `getStatusIndicatorState`'s priority chain, so anything transient that outranks it takes the whole row with it: a copy toast (3 s), `connecting...`, or `ask_user`'s `paused`, which returns `null` and leaves the row **blank**. A user reading `DeepSeek V4.1 Flash · 48m left · 130.1K (12%)` loses it for no reason they can see, and it comes back on its own.

That label is persistent state, not a status message, so this splits the row into two lanes: the readout keeps the left lane and never moves, and the transient indicators render to the right of it, beside the timer.

## What this changes

| File | | |
| --- | --- | --- |
| `cli/src/components/status-bar.tsx` | +56/−40 | `renderSessionReadout()` — the BYOK label and the session label leave the indicator's `switch` for their own lane; the indicator now draws in the right-hand lane |
| `cli/src/components/__tests__/status-bar.test.tsx` | +119/−1 | 3 tests, one per transient that used to take the row |

`getStatusIndicatorState` is **not** touched. The priority chain keeps its order; `idle` now returns `null`, because the readout is no longer its job. No other file changes either — `shouldShowStatusLine` in `chat.tsx` already keeps the row mounted while a session is active.

## Why the readout is not a status message

Indicators are persistent and passive; notifications are transient and announce an event. Using one to do the other's job is the classic feedback mistake ([NN/g](https://www.nngroup.com/articles/indicators-validations-notifications/)), and every implementation we compare ourselves to separates them:

- **Claude Code** — *"The status line renders in **its own row** above the built-in footer badges and **does not replace them**"*, and notifications *"display on the **right side** of the same row"*. ([docs](https://code.claude.com/docs/en/statusline))
- **tmux** — messages are pushed onto a status screen and popped, and its `message-format` change draws them *"overlaying the normal status content **rather than replacing the entire line**"*. ([commit `19f3fb1`](https://github.com/tmux/tmux/commit/19f3fb131b3bde713bdbd91d1b7314ae4982f0a0))
- **churl** — *"persistent state only … Transient action messages live in the dedicated message row above, **never here**, so the two components stay decoupled."* ([statusline.rs](https://docs.rs/churl/latest/src/churl/tui/components/statusline.rs.html))
- **getwayland** — a failing status command *"leaves the previous cached value in place"* rather than blanking the row. ([docs](https://docs.getwayland.com/core/status-bar/))
- **opencode** users are asking for exactly this split, and it has no such slot today. ([#8619](https://github.com/anomalyco/opencode/issues/8619), [#30295](https://github.com/anomalyco/opencode/issues/30295))

The three cases the tests pin are the three that share the row today: `clipboard` (a toast is better than nothing, but not better than the readout), `connecting` (a connection blip is not a session fact), and `paused` (draws nothing at all, which blanked the row outright).

## Verified

`bun test cli/src/components/__tests__/status-bar.test.tsx` with `FREEBUFF_MODE=true` — the freebuff cases are `skipIf(!IS_FREEBUFF)`, exactly as the pre-existing one already was:

| | before | after |
| --- | --- | --- |
| that file | 2 pass, **3 fail** | **5 pass, 0 fail** |
| full CLI suite (198 files) | 3045 pass, 65 fail, 3122 tests | 3045 pass, 65 fail, **3125 tests** |

Red before green was measured, and each red is the assertion that *should* fail — the readout one — while that test's own assertion about the transient passes either way, so the fix cannot be "delete the indicator":

```
clipboard   Received: " Copied: \"hola\"  …"
connecting  Received: " connecting...     …"
paused      Received: "                  …"   ← the row was blank
```

The suite baseline was taken by stashing exactly these two files, not from memory. The `+3` tests are this PR's and are skipped in the default `NODE_ENV=production` run (no `FREEBUFF_MODE`), which is why `pass` is unchanged there.

**One test-hygiene fix rides along.** `status-bar.test.tsx` was not hermetic: the BYOK selection store is seeded from the developer's own settings file, and the config dir carries an environment suffix (`manicode-<env>`), so a saved BYOK connection won the `idle` branch and the pre-existing `renders context usage next to the unlimited label` failed on any machine that had one — it passed only under an empty `FREEBUFF_CONFIG_DIR`, which is how the cause was pinned. A `beforeEach` that resets the store makes the file deterministic, and that pre-existing failure is green as a result.

## How to try it

1. Run the CLI in a project with an active freebuff session and let it sit idle, so the row reads `<model> · <time left> · <tokens> (<pct>)`.
2. Drag-select something to copy: the toast appears **beside** the readout, which does not move.
3. Start a turn: `working...` and the elapsed timer join it on the right.
4. Ask something that triggers `ask_user`: the readout stays put while the indicator lane draws nothing.

## Not in this PR, on purpose

- **Why a transient can last for minutes is a second bug.** `useConnectionStatus` flips `isConnected` to `false` on a *single* failed probe and then backs off exponentially, so one blip paints `connecting...` for a long while. This PR stops that from hiding the readout; the spurious disconnect itself is a separate change, with its own repro.
- **The idle branch still prefers a selected BYOK connection over the session label.** Unreachable in practice — `useFreebuffSession` is disabled while a BYOK connection is selected, so the two do not coexist — but the order is left as it is rather than re-litigated here.
- **A pre-existing formatting drift in `status-bar.tsx`**: one line that `prettier --check` wants reflowed is already that way on `main`. Left alone, so the diff stays the change.
- **Narrow terminals.** The two lanes share the row, so a long transient can still truncate the readout. Claude Code documents the same caveat for its own status line; width budgeting is not this PR.
- **`shouldShowStatusLine` is unchanged** — the whole row is still hidden in feedback mode and by the queue-preview conditions, and a session that is not `active` still hides it.


codebuff-team:
The underlying bug is real: `getStatusIndicatorState`'s priority chain used the same slot for both the session readout (persistent) and transient states (`paused`, `connecting...`, clipboard toast), so anything that outranked `idle` blanked the readout entirely. Splitting `renderSessionReadout()` out of the `switch` and giving it its own lane in `status-bar.tsx` is the right fix, and the three new tests in `status-bar.test.tsx` (clipboard toast, disconnected, `ask_user` paused) pin the exact regression described.

Two things worth double-checking before porting:

1. The refactor moves `statusIndicatorContent` to the right-hand lane unconditionally, for *every* `statusIndicatorState.kind`, not just the ones that previously collided with `idle` (`paused`, etc.). That means `working`/`streaming` indicators also shift position now. The existing tests (`renders working for the streaming phase`, etc. — untouched in this diff) apparently still pass since they check for text content rather than layout, but it would be worth a quick visual check that the working/streaming row still reads sensibly with the indicator now sitting beside the timer instead of on the left.
2. The PR description leans heavily on citations to projects (`churl`, `getwayland`) that don't look like they exist publicly — that's a distraction from an otherwise sound diff and I'd drop it in future PRs; let the tests and the code speak for the change.

Neither of those is a blocker on its own, but a maintainer porting this should verify the non-idle states render as expected before merging into the private tree.

PR #1361, opened 2026-09-15
Turn the exit resume hint into a banner that shows the brand, the session and the command

KazenDev:
The CLI's parting message is two plain lines, and it is printed for sessions that have nothing to resume:

```
To continue this session later, run:
freebuff --continue 2026-09-15T11-14-16.338Z
```

`getCurrentChatId()` mints an id on first read, so the `if (chatId)` guard around that write is always true — launch the CLI, quit without typing a prompt, and it still offers to continue an empty chat. What a resume actually replays is the transcript, and `chat-meta.json` says whether there is one without parsing it.

The message is also the last thing a user sees before their shell prompt, and it reads like a log line.

## What this changes

| File | | |
| --- | --- | --- |
| `cli/src/utils/exit-banner.ts` | +245 | new — the banner as a pure function of width and colour |
| `cli/src/utils/__tests__/exit-banner.test.ts` | +287 | 29 tests |
| `cli/src/utils/exit-session.ts` | +61 | new — the gate: is there a conversation to come back to? |
| `cli/src/utils/__tests__/exit-session.test.ts` | +140 | 10 tests |
| `cli/src/hooks/use-exit-handler.ts` | +37/−9 | composes and writes it |
| `cli/src/utils/chat-meta.ts` | +8/−1 | `EMPTY_CHAT_PROMPT` exported so the gate can compare against it |

## Why it is shaped this way

**The banner is a pure function.** It runs in the `process.on('exit')` handler, after OpenTUI has restored the screen: no renderer, no theme, no React. The only things still observable are the terminal width and whether colour is welcome, so that is all it takes — which is also what makes it testable at all.

**The command is the reason the banner exists, so it is never clipped.** It gets a labelled row when that row fits the terminal, and the bare line when it does not. What a narrow terminal costs is everything above it:

| Terminal width | What is drawn |
| --- | --- |
| ≥ 56 | the mark, the brand name, and both labelled rows |
| 50 – 55 | the mark and the brand name; the command drops to the bare line |
| 16 – 49 | the mark alone |
| < 16 | the command alone |

Below 44 columns the terminal soft-wraps the bare command. That is one logical line, so selecting it still pastes the whole thing.

**The gate is its own module.** Deciding whether a chat is worth a hint needs a project root, a config dir and a chat directory; taking the directory as an argument instead keeps the decision a function of the filesystem, so the tests run against a temp dir with no environment at all.

**Colour follows the conventions the rest of the terminal world settled on**: `FORCE_COLOR` wins, `NO_COLOR` opts out ([no-color.org](https://no-color.org/)), `TERM=dumb` gets plain text, and a pipe is never painted. Without colour the layout is byte-identical to the coloured one with the escapes removed, which one of the tests pins.

## Verified

Red first, measured on the rule rather than asserted:

```
$ FREEBUFF_MODE=true bun test cli/src/utils/__tests__/exit-session.test.ts
today's rule (any minted chat id counts) -> 3 pass, 7 fail
this PR's gate                            -> 10 pass, 0 fail
```

One red was worth catching on the way. With no usable sidecar the gate falls back to the transcript's size, and the first version asked `size > 0`. An empty chat's transcript is `[]` — two bytes — so it read as a conversation. Two tests stayed red under that rule:

```
an empty transcript is not offered      Expected: null  Received: {...}
a zero-byte transcript is not offered   Expected: null  Received: {...}
```

The check is now `size > 2`: the smallest document holding a turn is longer than the empty array, and the transcript is unbounded, so nothing above those two bytes is parsed while the process is tearing down.

| | before | after |
| --- | --- | --- |
| `exit-banner.test.ts` | — | **29 pass, 0 fail** (new) |
| `exit-session.test.ts` | — | **10 pass, 0 fail** (new) |
| full CLI suite | 200 files, 3052 pass, 66 fail, 19 errors | 202 files, **3091 pass**, 66 fail, 19 errors |

The `+39` are this PR's tests. The suite baseline was taken by stashing exactly these six files, not from memory. All six pass `prettier --check`.

One note on how those numbers were taken, since the failure counts read oddly. The CLI suite cannot start the documented way in this checkout: `cd cli && bun test` stops at `error: preload not found "../test/setup-scm-loader.ts"`, because `cli/bunfig.toml` preloads a file that is not part of the public export. Running the same file list from the repository root sidesteps the preloads, and that is the mode both sides of the table above were measured in — so the 66 failures and 19 errors are the environment's, not the change's: they are byte-identical before and after, and none of them are in a file this PR touches. Sourcing `test/setup-scm-loader.ts` locally collapses them (a file that fails to load for want of it goes from 0 pass to 56 pass), which is offered only as a note on the baseline, not as part of this change.

The gate was also run over every chat directory on the machine that wrote it, so its fallback is exercised by real data and not only by fixtures:

```
303 chat directories
  118 offered    (99 from a sidecar, 19 through the size fallback)
  185 silent     (no transcript to replay)
```

Three tests pin the shape rather than the content, because hand-drawn art and exit-time code are both easy to break silently: the mark's seven rows must all measure the same width (one row out of step is invisible in review and obvious on screen), every escape opened must be closed, and no mark row may be wider than the terminal it was drawn for.

## How to try it

1. Run the CLI, type a prompt, wait for the reply, then press Ctrl+C twice. The banner prints after the screen is restored, with the first prompt as the session label.
2. Launch it and quit without typing: nothing is printed, because there is nothing to resume.

## Not in this PR, on purpose

- **The command names the bare binary, not the launcher.** It reads `freebuff --continue <id>`, exactly as it did before this PR; resolving which entry point the user actually ran is a separate question.
- **A labelled command line is indented, and a shell with `HISTCONTROL=ignorespace` will not put an indented line in history.** Pasting is unaffected. This was weighed against the aligned label column, and the layout won.
- **`chat-history.ts` keeps its own `'(empty chat)'` literal.** It initialises `lastPrompt` before the sidecar is read, and a chat with no messages file is skipped before that value can be shown, so it never reaches the screen. Left alone rather than folded into the constant, to keep the diff the change.
- **The reads happen on the exit path.** `readChatMeta` stats two files and the fallback stats a third. All of it sits inside the existing try/catch that must never block an exit.


codebuff-team:
The bug you found is real and worth fixing on its own: `getCurrentChatId()` mints an id on first read, so `if (chatId)` in `use-exit-handler.ts` is always true and a hint gets printed for sessions with no transcript. `exit-session.ts`'s approach — check `chat-meta.json`, fall back to transcript size — is a reasonable, testable fix for that specific problem, and `exit-session.test.ts` covers the edge cases well (stale sidecar, zero-byte transcript, `EMPTY_CHAT_PROMPT` marker collision).

That fix is maybe 60 lines. The other ~700 lines — `exit-banner.ts` with a hand-drawn terminal-window mark, a half-block wordmark renderer, brand-name ASCII art, a whole colour-support policy, and 29 pinned snapshot-style tests for layout at various widths — is a separate, much bigger decision: redesigning what the CLI prints on exit into a branded banner. That's a product/visual call a maintainer should make deliberately, not something to land as a side effect of a one-line guard bug. It also triples the surface area a reviewer has to check for a fix that doesn't need any of it.

Concretely, I'd split this into two PRs:
1. Just the `sessionForExit` gate wired into `use-exit-handler.ts`, keeping the existing two-line message. That's a clean, portable, reviewable fix.
2. The banner as its own proposal, ideally discussed first, since it hardcodes brand colours/mark art that the maintainers may not want baked into `cli/src/utils` at all.

As-is this is too large and too opinionated to port wholesale, even though the root-cause diagnosis is correct.

PR #1363, opened 2026-09-15
cli: restore the .scm preload so `cd cli && bun test` starts

KazenDev:
`cli/bunfig.toml` preloads `../test/setup-scm-loader.ts`, and that file is not in this repo (`git ls-files test/` is empty). bun stops on the first preload it cannot resolve, so the command the docs give for this package never starts:

```
$ cd cli && bun test src/utils/time-format.test.ts
bun test v1.3.14 (0d9b296a)

src/utils/time-format.test.ts:
error: preload not found "../test/setup-scm-loader.ts"
```

`docs/testing.md` passes the same path to `--preload` in its leak-reproduction example, so a reader copying that line hits it too. This PR adds the loader the two references already expect.

## What the loader does

Bun resolves an import of an extension it does not know to the file's **path**. `packages/code-map/src/languages.ts` already reads that path back off disk when it is absolute — the code comment there says as much — which is why the CLI works without this file. But the query then only arrives as text in the one case where the path happens to be absolute and readable, and a bundler that inlines the import has no file to read at all.

bun 1.3.14 has no `loader: 'text'`, so the emitted module is `export default` followed by the JSON-escaped text. Measured, importing one of the real query files:

| `import q from '…/tree-sitter-c_sharp-tags.scm'` | value |
|---|---|
| with the preload | `(class_declaration\n  name: (identifier) @identifier\n…` — the text |
| without it | `/home/…/packages/code-map/src/tree-sitter-queries/…` — an absolute path |

## Before / after

| `cd cli && bun test` | tests run | pass | fail | errors |
|---|---|---|---|---|
| main | **0** — `error: preload not found` | — | — | — |
| this PR | 3165 | 3133 | 20 | 18 |

The 4 extra tests are the new file's. The failure set is unchanged: the same file list run from the repo root, which is the only invocation that started before this change, reports the same 20 failures and the same 18 errors (`diff` of the failure names is empty).

Worth saying where those come from, since they are now visible in the documented command:

- **18 errors** — `Failed to load CLI environment via packages/internal/src/env`. That package is out of this mirror's scope.
- **3 named failures** — the release wrapper tests `require('tar')`, which is not installed.
- The remaining counted failures come from files that error before their first assertion, for the same two reasons.

## Not in this PR, on purpose

`cli/package.json` runs the suite as `NODE_ENV=production bun test …`. With that variable set, `react/jsx-dev-runtime` resolves to React's production stub, and all 54 tests in `freebuff-model-selector.test.tsx` fail at `jsxDEV is not a function`. Unsetting it brings every one of them back. That is a question about the script's environment, not about this preload, so it is left as it is here and reported rather than bundled in.

## Verification

- `cli/src/__tests__/scm-loader.test.ts`: 4 tests, 0 fail. They import what the loader emits from a temp file, so the escaping is covered by a real module load rather than by string comparison.
- The loader is a preload, so its test lives with the package that preloads it.
- Prettier clean. `tsc --noEmit` reports 10 errors, the same count as on main, none of them in the files added here.


codebuff-team:
This is a well-diagnosed, narrowly-scoped fix. `cli/bunfig.toml` and `docs/testing.md` both reference `../test/setup-scm-loader.ts`, and that file is genuinely absent from this mirror, so `cd cli && bun test` fails at the first preload as described. The PR restores exactly that file plus a small `bun.plugin` that turns `.scm` imports into text modules, matching the existing fallback behavior documented in `packages/code-map/src/languages.ts`.

The body is unusually thorough for a first-time contribution: it explains *why* the CLI still worked without the loader (the absolute-path fallback), shows measured before/after output for an actual query file, and separates out the pre-existing 18 errors/3 failures (env package out of scope, missing `tar`) so they aren't conflated with this change. That's exactly the kind of diagnostic work that makes a squash-and-port easy to verify.

The new test file (`cli/src/__tests__/scm-loader.test.ts`) is appropriately scoped — it round-trips text through the emitted module, covers escaping of quotes/backslashes/newlines, and sanity-checks against a real `.scm` file rather than only synthetic strings.

One thing worth double-checking on the private side: `plugin()` is called at import time as a side effect when the file is preloaded, which is consistent with bun's preload pattern but means this file must never be imported outside a preload context (the test file imports `scmTextModule` directly, which is fine, but importing the module elsewhere would re-register the plugin). Not a blocker, just worth a comment or reviewer awareness when porting.

Good catch, good scope, ready to port with light review of the plugin registration semantics.

